PR #54, opened 2017-12-03
Additional render options

jdent02:
Bounce counts, volume settings, and lighting sampler

Note: export only option is partially complete so it isn't exposed in the ui yet.

### DIFF
--- a/properties/scene.py
+++ b/properties/scene.py
@@ -156,7 +156,7 @@ class AppleseedRenderSettings(bpy.types.PropertyGroup):
         cls.light_sampler = bpy.props.EnumProperty(name="Light Sampler",
                                                   description="The method used for sampling lights",
                                                   items=[('cdf', 'CDF', 'CDF'),
-                                                        ('lighttree', 'Lights Tree', 'Light Tree')],
+                                                        ('lighttree', 'Light Tree', 'Light Tree')],
                                                   default='cdf')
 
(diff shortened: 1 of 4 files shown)


properties/scene.py
```diff
@@ -147,14 +153,20 @@ def register(cls):
                                                     min=1,
                                                     max=999999)
 
+        cls.light_sampler = bpy.props.EnumProperty(name="Light Sampler",
+                                                  description="The method used for sampling lights",
+                                                  items=[('cdf', 'CDF', 'CDF'),
+                                                        ('lighttree', 'Lights Tree', 'Light Tree')],
```
dictoon: `Lights Tree` -> `Light Tree`

properties/scene.py
```diff
@@ -198,6 +210,34 @@ def register(cls):
                                                 min=0,
                                                 max=512)
 
+        cls.use_separate_bounces = bpy.props.BoolProperty(name="Use Individual Bounce Limits",
+                                                          description="Use individual limits for different ray types",
+                                                          default=False)
+
+        cls.max_diffuse_bounces = bpy.props.IntProperty(name="Max Diffuse Bounces",
+                                                        description="Maximum diffuse length: 0 = Unlimited",
```
dictoon: This would be more accurate: `Maximum total number of diffuse bounces: 0 = unlimited` (same for the other bounce types).

jdent02:
Done....

dictoon:
Merged, thank you!

PR #93, opened 2017-12-05
Add auto threads option

jdent02:
Ignore all those commits where I was fixing the merge from master that I jacked up......


properties/scene.py
```diff
@@ -74,6 +74,10 @@ def register(cls):
                                                         ('export_only', 'Export Files Only', '')],
                                                  default='render')
 
+        cls.threads_auto = bpy.props.BoolProperty(name="Auto Threads",
+                                                description="Automatically determine the number of rendering threads",
```
dictoon: Let's align arguments vertically so that PyCharm stops complaining :)

PR #96, opened 2017-12-07
ui clean-up/update 01

vallmoy:


properties/world.py
```diff
@@ -83,55 +78,59 @@ def register(cls):
...
+                                                min=0.0,
+                                                max=10.0)
 
-        cls.horiz_shift = bpy.props.FloatProperty(name="Horizon Shift",
-                                                  description='',
-                                                  default=0.0,
-                                                  min=-2.0,
-                                                  max=2.0)
+        cls.turbidity_multiplier = bpy.props.FloatProperty(name="Turbidity Multiplier",
+                                                    description='',
```
dictoon: I suggest vertically aligning arguments so that PyCharm stops complaining :) Here and below as well.

properties/materials.py
```diff
@@ -843,6 +843,16 @@ class AppleseedMatProps(bpy.types.PropertyGroup):
         name="Light Near Start", description="Amount by which to extend the start of light's influence away from the emissive material", default=0.0,
         min=0, max=10, update=refresh_preview)
 
+    export_emitting_obj_as_lights = bpy.props.BoolProperty(name="Export As Mesh Light",
+                                                                   description="Export object with light-emitting materials as mesh (area) lights",
```
dictoon: *object* -> *objects* (the typo had always been there)

properties/materials.py
```diff
@@ -847,7 +847,7 @@ class AppleseedMatProps(bpy.types.PropertyGroup):
                                                                    description="Export object with light-emitting materials as mesh (area) lights",
                                                                    default=True)
 
-    light_mats_radiance_multiplier = bpy.props.FloatProperty(name="Radiance Multiplier",
+    light_mats_radiance_multiplier = bpy.props.FloatProperty(name="Exposure",
```
dictoon: That's not an exposure, just a linear multiplier. Exposure is a specific control (an exponential multiplier) that we'll want to expose as well.

ui/materials.py
```diff
@@ -985,8 +985,28 @@ def draw(self, context):
...
+            # col = split.column()
+            # col.prop(asr_mat, "material_use_alpha", text="Alpha Map", icon="POTATO", toggle=True)
+            # col = split.column()
+            # if asr_mat.material_use_alpha:
+                # col.prop_search(asr_mat, "material_alpha_map", material, "texture_slots", text="")
+                # if asr_mat.material_alpha_map != '':
+                    # alpha_tex = bpy.data.textures[asr_mat.material_alpha_map]
+                    # layout.prop(alpha_tex.image.colorspace_settings, "name", text="Color Space")    
+
+            # layout.prop(asr_mat, "shade_alpha_cutouts")
```
dictoon: The "Shade Alpha Cutouts" feature has been removed from appleseed so it can go.

ui/materials.py
```diff
@@ -985,8 +985,28 @@ def draw(self, context):
...
             #----------------------------------------------
 
-            # bump/normal mapping
+            # alpha
+            
             layout.separator()
+
+            layout.prop(asr_mat, "material_alpha")
+
+            # split = layout.split(percentage=0.50)
```
dictoon: What's the deal with this commented code? Should it be uncommented, or removed?

properties/scene.py
```diff
@@ -106,7 +106,7 @@ def register(cls):
                                                         description='Avoid correlation patterns at the expense of slightly more sampling noise',
                                                         default=True)
 
-        cls.pixel_filter = bpy.props.EnumProperty(name="Pixel Filter",
+        cls.pixel_filter = bpy.props.EnumProperty(name="Filter",
```
dictoon: I had done this change recently (adding "Pixel" in front of "Filter" and "Filter Size") to make it clear these were controlling pixel filtering (there are many filtering in a renderer).

That said, from a user's point of view maybe the terms "Filter" and "Filter Size" are completely unambiguous in which case it's fine to omit "Pixel".

Ultimately I let you decide, I just wanted to point your attention to this :)

ui/render.py
```diff
@@ -224,11 +224,12 @@ def draw(self, context):
 
         layout.prop(asr_scene_props, "shutter_open")
         layout.prop(asr_scene_props, "shutter_close")
+        # layout.prop(asr_scene_props, "mblur_samples")
```
dictoon: Remove? We try not to leave any dead code in the repo.

dictoon:
Looks awesome, thanks!

properties/scene.py
vallmoy: I changed it only because i want blenderseed names to match appleseed.studio names

ui/materials.py
vallmoy: Fixed, i didn't know if cutouts are removed or missing

properties/materials.py
vallmoy: Fixed

properties/materials.py
vallmoy: Done

properties/world.py
vallmoy: Fixed

ui/render.py
vallmoy: Removed

ui/materials.py
vallmoy: Alpha map is broken, so i disabled it from ui

dictoon:
Superseded by #99.

PR #97, opened 2017-12-07
Multiple items

jdent02:
Plastic BRDF
Material ui redo
Material property cleanup

### DIFF
--- a/ui/materials.py
+++ b/ui/materials.py
@@ -165,14 +165,19 @@ class AppleseedMaterialShading(bpy.types.Panel):
                         layout.prop(mix_tex.image.colorspace_settings, "name", text="Color Space")
 
                     # reflectance
-                    split = layout.split(percentage=0.90)
+                    split = layout.split(percentage=0.40)
+                    col = split.column()
+                    col.label("Reflectance")
+                    split = split.split(percentage=0.83)
                     col = split.column()
                     col.prop(current_layer, "lambertian_reflectance", text="")
+
                     if current_layer.lambertian_use_diff_tex:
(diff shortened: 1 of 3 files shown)


ui/materials.py
```diff
@@ -228,33 +237,41 @@ def draw(self, context):
...
                     if current_layer.glass_reflection_tint_use_tex:
                         layout.prop_search(current_layer, "glass_reflection_tint_tex", material, "texture_slots")
 
+                    split = split.split(percentage=1.0)
                     col = split.column()
                     col.prop(current_layer, "glass_reflection_tint_use_tex", text="", icon="TEXTURE_SHADED", toggle=True)
                     if current_layer.glass_reflection_tint_tex != '' and current_layer.glass_reflection_tint_use_tex:
-                        glass_reflection_tint_tex = bpy.data.textures[current_layer.glass_reflection_tint_tex]
-                        layout.prop(glass_reflection_tint_tex.image.colorspace_settings, "name", text="Color Space")
+                        specular_tex = bpy.data.textures[current_layer.glass_reflection_tint_tex]
```
dictoon: I think this variable name is wrong (it's reused in many places). Not that it really makes a difference, but until we refactor all this code to reduce duplication, we should probably use correct names.

dictoon: I propose to fix this after the fact. I'll merge the PR right away so we can start testing.

jdent02: Yeah I’m doing a lot of copy/paste for the BSDF additions.  All of the actual property calls are updated but sometimes I miss the variables inside the functions.

PR #99, opened 2017-12-07
UI cleanups and fixes by @vallmoy

dictoon:


jdent02:
Bit late to the party but this all looked fine to me.

dictoon:
> Bit late to the party but this all looked fine to me.

Appreciated nevertheless!

Issue #58, opened 2017-12-04
Add tile size


vallmoy:
Blenderseed uses 64x64 tile by default

PR #121, opened 2017-12-13
Add tile size

jdent02:


jdent02:
Went ahead and did this since luke's been asking for it.

Closes #58

projectwriter.py
```diff
@@ -2598,9 +2598,11 @@ def __emit_output(self, scene):
     def __emit_frame_element(self, scene):
         camera = scene.camera
         width, height = self.__get_frame_resolution(scene.render)
+        tile_size = str(scene.appleseed.tile_size_x) + " " + str(scene.appleseed.tile_size_y)
         self.__open_element("frame name=\"beauty\"")
         self.__emit_parameter("camera", "camera" if camera is None else camera.name)
         self.__emit_parameter("resolution", "{0} {1}".format(width, height))
+        self.__emit_parameter("tile_size", tile_size)
```
dictoon: I suggest getting rid of the temporary variable and using `"".format()` like it's done just above.

properties/scene.py
```diff
@@ -90,6 +90,16 @@ def register(cls):
                                                         description='Avoid correlation patterns at the expense of slightly more sampling noise',
                                                         default=True)
 
+        cls.tile_size_x = bpy.props.IntProperty(name="tile_size_x",
+                                                description="Set the x size of the render tile",
```
dictoon: Replace *x size* by *width* (also in the property name and the variable). Similarly, *y size* -> *height*.

jdent02:
Done

PR #125, opened 2017-12-15
Environment improvements

jdent02:
Env alpha

Additional latlong parameters

### DIFF
--- a/properties/world.py
+++ b/properties/world.py
@@ -126,6 +126,11 @@ class AppleseedSkySettings(bpy.types.PropertyGroup):
                                                description="Texture to influence environment",
                                                default="")
 
+        cls.env_alpha = bpy.props.FloatProperty(name="env_alpha",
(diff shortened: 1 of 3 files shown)


properties/world.py
```diff
@@ -126,6 +126,28 @@ def register(cls):
...
+        cls.horizontal_shift = bpy.props.FloatProperty(name="horizontal_shift",
+                                                       description="Environment texture horizontal shift in degrees",
+                                                       default=0.0,
+                                                       min=-360.0,
+                                                       max=360.0)
+
+        cls.vertical_shift = bpy.props.FloatProperty(name="vertical_shift",
+                                                     description="Environment texture vertical shift in degrees",
+                                                     default=0.0,
+                                                     min=-360.0,
```
dictoon: Vertical range should be -180.0 to +180.0.

PR #126, opened 2017-12-17
Add orthographic camera

jdent02:


projectwriter.py
```diff
@@ -2292,23 +2293,33 @@ def __emit_camera(self, scene):
         render = scene.render
 
         film_width = camera.data.sensor_width / 1000
+        ortho_width = camera.data.ortho_scale
         aspect_ratio = self.__get_frame_aspect_ratio(render)
-        lens_unit = "focal_length" if camera.data.lens_unit == 'MILLIMETERS' else "horizontal_fov"
+        # lens_unit = "focal_length" if camera.data.lens_unit == 'MILLIMETERS' else "horizontal_fov"
```
dictoon: You can remove the dead code.

projectwriter.py
```diff
@@ -2292,23 +2293,33 @@ def __emit_camera(self, scene):
...
 
         # Blender's camera focal length is expressed in mm
-        focal_length = camera.data.lens / 1000.0
+        # focal_length = camera.data.lens / 1000.0
```
dictoon: Remove dead code here too.

projectwriter.py
```diff
@@ -2283,6 +2283,7 @@ def __emit_camera(self, scene):
         width = scene.render.resolution_x
         height = scene.render.resolution_y
         emit_diaphragm_map = False
+        asr_cam = camera.data.appleseed
```
dictoon: Let's go for `appleseed_cam` if we need a shorter alias for `camera.data.appleseed`. The `asr` prefix is a shorthand for the `appleseed::renderer` C++ namespace.

PR #127, opened 2017-12-17
Fix for auto threads

jdent02:


render.py
```diff
@@ -189,11 +189,12 @@ def __render_project_file(self, scene, project_filepath):
             max_x = width - 1
             max_y = height - 1
 
-        # Launch appleseed.cli.
+        # Launch appleseed.cli.\
```
dictoon: Remove trailing backslash.

Issue #47, opened 2017-11-28
Rendering progress indicator doesn't change?


dictoon:

jdent02:
#130

PR #130, opened 2017-12-19
Fix for progress bar update

jdent02:


jdent02:
Closes #47

render.py
```diff
@@ -234,6 +240,7 @@ def __render_project_file(self, scene, project_filepath):
         # Make sure the appleseed.cli process has terminated.
         process.kill()
 
+
```
dictoon: Remove blank line. PEP8 mandate one blank line between methods (but two between free functions).

render.py
```diff
@@ -177,6 +177,12 @@ def __render_project_file(self, scene, project_filepath):
         # Compute render resolution.
         (width, height) = util.get_render_resolution(scene)
 
+        # Compute total pixel count
```
dictoon: End full line comments with a period.

render.py
```diff
@@ -292,6 +299,13 @@ def __process_tile_data_chunk(self, process, min_x, min_y, max_x, max_y):
         layer.rect = pix
         self.end_result(result)
 
+        # Update progress bar
+        tile_pixels = tile_w * tile_h
+        self.rendered_pixels += tile_pixels
+        percentage = float(self.rendered_pixels / self.total_pixels)
```
dictoon: Not quite a percentage, maybe just `progress`?

Why is the `float()` part necessary?

render.py
```diff
@@ -292,6 +299,13 @@ def __process_tile_data_chunk(self, process, min_x, min_y, max_x, max_y):
...
+        self.update_progress(percentage)
+
```
dictoon: Remove one blank line.

render.py
```diff
@@ -292,6 +299,13 @@ def __process_tile_data_chunk(self, process, min_x, min_y, max_x, max_y):
         layer.rect = pix
         self.end_result(result)
 
+        # Update progress bar
+        tile_pixels = tile_w * tile_h
```
dictoon: You can get rid of the intermediary variable `tile_pixels`.

render.py
```diff
@@ -177,6 +177,12 @@ def __render_project_file(self, scene, project_filepath):
         # Compute render resolution.
         (width, height) = util.get_render_resolution(scene)
 
+        # Compute total pixel count
+        self.total_pixels = width * height * scene.appleseed.renderer_passes
```
dictoon: You need to take into account the render window: nstead of `width` and `height`, you probably need to use `max_x - min_x + 1` and `max_y - min_y + 1`.

render.py
dictoon: Similarly here, you need to take the render window into account: `take_x` and `take_y` seems to be what you're after.

jdent02:
Fixes #134 #139 #47

properties/materials.py
```diff
@@ -1096,6 +1085,23 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
 
+    orennayar_brdf_reflectance_multiplier = bpy.props.FloatProperty(name="orennayar_brdf_reflectance_multiplier",
+                                                                    description="Reflectance multiplier",
+                                                                    default=1.0,
+                                                                    min=0.0,
+                                                                    max=1.0,
+                                                                    update=refresh_preview)
+
+    orennayar_brdf_use_reflect_multiplier_tex = bpy.props.BoolProperty(name="orennayar_brdf_use_rough_tex",
+                                                          description="Use a texture to influence roughness",
```
dictoon: Indent.

properties/materials.py
```diff
@@ -1096,6 +1085,23 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
+                                                          default=False,
+                                                          update=refresh_preview)
+
+    orennayar_brdf_reflect_multiplier_tex = bpy.props.StringProperty(name="orennayar_brdf_rough_tex",
+                                                        description="Roughness texture",
```
dictoon: Indent.

properties/materials.py
```diff
@@ -1558,6 +1564,13 @@ class AppleseedMatProps(bpy.types.PropertyGroup):
                                                       max=1.0,
                                                       update=refresh_preview)
 
+    material_bump_offset = bpy.props.FloatProperty(name="material_bump_offset",
+                                                   description="Offset of bump map values",
```
dictoon: Rather: "Offset value to evaluate bump mapping".

properties/nodes/__init__.py
```diff
@@ -46,6 +46,16 @@ def poll(cls, context):
         renderer = context.scene.render.engine
         return renderer == 'APPLESEED_RENDER'
 
+    def update(self):
+        self.refresh = True
+
+    def acknowledge_change(self, context):
+        while self.refresh:
```
dictoon: Is this a known / documented pattern? Why not simply set `self.refresh` to False? Is `self.refresh` modified by other threads concurrently?

jdent02: I borrowed it from the Lux exporter.  I tried several other methods but this was the only one I could get to work.

dictoon: I'm actually very surprised that it has a different behavior from a simple

    self.refresh = False

As far as I can tell, both have the exact same semantic. Are you sure?

dictoon: A _strictly identical_ construction would be

    if self.refresh:
        self.refresh = False

If the simple version doesn't work, could you try this?

jdent02: Simple version works

Issue #111, opened 2017-12-11
Expose Low Light Threshold parameter


dictoon:
The property `dl_low_light_threshold` already exists, it's simply not exposed in the UI yet.

Issue #72, opened 2017-12-04
Add size and distance to sun lamp


vallmoy:
Size multiplier and distance are missing

Issue #68, opened 2017-12-04
Add BSDF nodes


vallmoy:
BSDF nodes that are currently missing:
- Blinn BRDF
- Glass BSDF
- Glossy BRDF
- Metal BRDF
- Plastic BRDF
- Sheen BRDF

dictoon:
Metal BRDF does not work, the following message appears in the system console:
```
error    : Could not open C:\Users\Franz\AppData\Local\Temp\mat_preview\mat_preview.appleseed for writing
```

PR #144, opened 2017-12-24
Add missing nodes/other stuff

jdent02:
#68 
#144 
#111 
#72 
#71 
#49


dictoon:
That's awesome, thanks Jon!

It would be best to do one commit per feature or bug fix. For instance, #111, #72, #71 and #49 each deserve a separate commit. Right now all we will see in the repo for all this work is a single commit with title "Add plastic BRDF node".

Would you mind trying to split up this work into multiple commits?

PR #157, opened 2018-01-01
Small comment for node refresh

jdent02:


properties/nodes/__init__.py
```diff
@@ -46,6 +46,7 @@ def poll(cls, context):
         renderer = context.scene.render.engine
         return renderer == 'APPLESEED_RENDER'
 
+    # This following code is required to avoid a max recursion error when Blender checks for node updates
```
dictoon: I wouldn't have put the comment here, but just above the `while` in `acknowledge_connection`, which really is the curious part (since the `while` is clearly completely redundant in pure Python).

Issue #70, opened 2017-12-04
Add volume node


vallmoy:
Generic volume node is missing

Issue #69, opened 2017-12-04
Add BSSRDF nodes


vallmoy:
BSSRDF nodes that are currently missing:
- Better Dipole BSSRDF
- Directional Dipole BSSRDF
- Gaussian BSSRDF
- Normalized Diffusion BSSRDF
- Standard Dipole BSSRDF

Issue #64, opened 2017-12-04
Add volume model


vallmoy:
Generic volume model is missing

Issue #63, opened 2017-12-04
Add BSSRDF models


vallmoy:
BSSRDF models that are currently missing:
- Better Dipole BSSRDF
- Directional Dipole BSSRDF
- Gaussian BSSRDF
- Normalized Diffusion BSSRDF
- Standard Dipole BSSRDF

PR #158, opened 2018-01-04
Add remaining materials and object flags

jdent02:


projectwriter.py
```diff
@@ -1018,33 +1033,38 @@ def __emit_back_material_bsdf_tree(self, material, material_name, scene, layers,
     # -----------------------------
     def __emit_bsdf_mixes(self, bsdfs):
 
+        # No BSDF
+        if not bsdfs:
```
dictoon: Why is this method called with zero BSDF?

properties/materials.py
```diff
@@ -67,13 +67,8 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
                                               ('specular_brdf', "Specular BRDF", ""),
-                                              ('specular_btdf', "Specular BTDF", "")
-                                              # ('better_dipole_bssrdf', "Better Dipole BSSRDF", ""),
-                                              # ('directional_dipole_bssrdf', "Directional Dipole BSSRDF", ""),
-                                              # ('gaussian_bssrdf', "Gaussian BSSRDF", ""),
-                                              # ('normalized_diffusion_bssrdf', "Normalized Diffusion BSSRDF", ""),
-                                              # ('standard_dipole_bssrdf', "Standard Dipole BSSRDF", "")
-                                              ],
+                                              ('specular_btdf', "Specular BTDF", ""),
+                                              ('bssrdf', "Subsurface scattering BSSRDF", "")],
```
dictoon: Hm, "Subsurface scattering BSSRDF" is not a BSDF type. Why do we need to have it here?

properties/materials.py
```diff
@@ -1477,6 +1472,119 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
+                                            max=1.0,
+                                            update=refresh_preview)
+
+    bssrdf_weight_use_texture = bpy.props.BoolProperty(name="bssrdf_weight_use_texture",
+                                                       description="Use a texture for the BSSRDF weight",
+                                                       default=False,
+                                                       update=refresh_preview)
+
+    bssrdf_weight_texture = bpy.props.StringProperty(name="bssrdf_weight_texture",
+                                                     description="Texture to influence bssrdf weight",
```
dictoon: *BSSRDF* (everywhere)

properties/materials.py
```diff
@@ -1477,6 +1472,119 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
+    bssrdf_mfp = bpy.props.FloatVectorProperty(name="bssrdf_mfp",
+                                               description="Mean Free Path",
+                                               default=(0.5, 0.5, 0.5),
+                                               subtype='COLOR',
+                                               min=0.0,
+                                               max=1.0,
+                                               update=refresh_preview)
+
+    bssrdf_mfp_use_texture = bpy.props.BoolProperty(name="bssrdf_mfp_use_texture",
+                                                    description="Use a texture for the BSSRDF mfp",
```
dictoon: Maybe "Use a texture for the Mean Free Path"?

properties/materials.py
```diff
@@ -1477,6 +1472,119 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
+                                                    default=False,
+                                                    update=refresh_preview)
+
+    bssrdf_mfp_texture = bpy.props.StringProperty(name="bssrdf_mfp_texture",
+                                                  description="Texture to influence bssrdf mfp",
```
dictoon: Maybe "Texture to influence Mean Free Path"?

properties/materials.py
```diff
@@ -1477,6 +1472,119 @@ class AppleseedMatLayerProps(bpy.types.PropertyGroup):
...
+
+    bssrdf_mfp_multiplier = bpy.props.FloatProperty(name="bssrdf_mfp_multiplier",
+                                                    description="Mean Free Path Multiplier",
+                                                    default=1.0,
+                                                    min=0.0,
+                                                    max=1.0,
+                                                    update=refresh_preview)
+
+    bssrdf_mfp_multiplier_use_texture = bpy.props.BoolProperty(name="bssrdf_mfp_multiplier_use_texture",
+                                                               description="Use a texture for the BSSRDF mfp multiplier",
```
dictoon: Maybe "Use a texture for the Mean Free Path multiplier"?

projectwriter.py
jdent02: Without it a default BSDF is always exported even if the material only uses SSS or volumes.

projectwriter.py
```diff
@@ -2125,6 +2131,39 @@ def __emit_bssrdf(self, material, bsdf_name, scene, layer=None, node=None):
         self.__emit_parameter("fresnel_weight", bssrdf_fresnel_weight)
         self.__close_element("bssrdf")
 
+    # ----------------------
+    # Write BSSRDF.
```
dictoon: Fix comment.

properties/materials.py
jdent02: Because it would require a redo of the material UI to include separate sections for both SSS and volume materials.  I suppose we can do that if needed but I was working under the assumption that we wanted to keep every material type under a single list.

export.py
```diff
@@ -44,14 +47,34 @@ class ExportAppleseedScene(bpy.types.Operator, ExportHelper):
     filename_ext = ".appleseed"
     filter_glob = bpy.props.StringProperty(default="*.appleseed", options={'HIDDEN'})
 
+    compress_export = bpy.props.BoolProperty(name="Create compressed archive",
+                                            description="Compress export (including all textures) into archive file",
```
dictoon: Fix indentation.

export.py
```diff
@@ -44,14 +47,34 @@ class ExportAppleseedScene(bpy.types.Operator, ExportHelper):
...
 
     def execute(self, context):
+        export_path = util.realpath(self.filepath)
         writer = projectwriter.Writer()
-        writer.write(context.scene, util.realpath(self.filepath))
+        writer.write(context.scene, export_path)
+
+        if self.compress_export:
+            appleseed_bin_dir = bpy.context.user_preferences.addons['blenderseed'].preferences.appleseed_binary_directory
+            appleseed_bin_path = os.path.join(appleseed_bin_dir, "projecttool")
```
dictoon: Rename `appleseed_bin_path` to `projecttool_path`.

projectwriter.py
```diff
@@ -676,6 +679,23 @@ def __emit_object_instance_element(self, object_name, instance_name, instance_ma
         self.__emit_line('<assign_material slot="0" side="back" material="{0}" />'.format(back_material_name))
         self.__close_element("object_instance")
 
+    # --------------------------------------------
+    # Emit object visibility flags
```
dictoon: Let's start using docstrings (triple quotes), and converting the existing block comments into docstrings.

properties/materials.py
dictoon: Wouldn't it be more accurate to have an item that says "No BSDF" then? Since that's how the appleseed material will end up looking?

jdent02: Possibly, but then if the user wants to use SSS or volume they have to add an additional material layer and change the default to 'no BSDF'.  That's two additional steps.

Right now if they want a volume they can just hit 'new material' and change the default layer to a volume and be happy without any additional steps.

jdent02:
#161 
#64 #70 
#63 #69 
#113 
#83

properties/nodes/material.py
```diff
@@ -89,6 +89,27 @@ def get_socket_value(self,texture_only=True):
...
+
+    def draw(self, context, layout, node, text):
+        layout.label(text)
+
+    def draw_color(self, context, node):
+        return 0.0, 0.8, 0.0, 1.0
+
+    def get_socket_value(self,texture_only=True):
+        if self.is_linked:
+                linked_node = self.links[0].from_node
```
dictoon: Fix indentation.

properties/nodes/volume.py
```diff
@@ -0,0 +1,111 @@
...
+#
+
+import bpy
+from bpy.types import Node
+from ...util import asUpdate
+from . import AppleseedNode
+
+
+class AppleseedVolumeNode(Node, AppleseedNode):
+    """appleseed Metal BRDF Node"""
```
dictoon: Fix docstring.

properties/nodes/volume.py
```diff
@@ -0,0 +1,111 @@
...
+        self.outputs.new('NodeSocketShader', "Volume")
+
+    def draw_buttons(self, context, layout):
+        layout.prop(self, "volume_absorption", text="Absorption")
+        layout.prop(self, "volume_absorption_multiplier", text="Absorption Multiplier")
+        layout.prop(self, "volume_scattering", text="Scattering")
+        layout.prop(self, "volume_scattering_multiplier", text="Scattering Multiplier")
+        layout.prop(self, "volume_phase_function_model", text="Phase Function Model")
+        if self.volume_phase_function_model == 'henyey':
+            layout.prop(self, "volume_average_cosine", text="Avergage Cosine")
```
dictoon: Fix typo.

properties/nodes/volume.py
jdent02: According to PEP8 single line docstrings are supposed to have the end quotes on the same line.  Or were you referring to something else?

jdent02:
Everything should be fixed up.  I tried tackling the block comment to docstring thing but you'll have to let me know if it's correct.

properties/nodes/volume.py
dictoon: The docstring says "Metal BRDF Node" but it seems like it should say "Volume Node" :)

projectwriter.py
```diff
@@ -752,51 +725,55 @@ def __emit_material(self, material, scene):
 
         return front_material_name, back_material_name
 
-    # Emit front material.
     def __emit_front_material(self, material, material_name, scene, layers, material_node=None, node_list=None):
-        # material_name here is material.name + "_front"
-        bsdf_name = self.__emit_front_material_bsdf_tree(material, material_name, scene, layers, material_node, node_list)
+        """Material_name here is material.name + "_front"""
```
dictoon: There's a missing closing quote here, maybe it'll be clearer this way:

    """Material_name here is material.name + _front"""

projectwriter.py
```diff
@@ -2188,19 +2259,21 @@ def __emit_material_element(self, material_name, bsdf_name, edf_name, surface_sh
                     bump_map += "_inst"
                     method = "normal" if asr_mat.material_use_normalmap else "bump"
 
-                if asr_mat.material_use_alpha and asr_mat.material_alpha_map != "":
-                    material_alpha_map = asr_mat.material_alpha_map + "_inst"
-                    self.__emit_texture(bpy.data.textures[asr_mat.material_alpha_map], False, scene)
-                else:
-                    material_alpha_map = asr_mat.material_alpha
+                # if asr_mat.material_use_alpha and asr_mat.material_alpha_map != "":
```
dictoon: What's the deal with this code? Should it be removed?

properties/nodes/volume.py
jdent02: Oh that’s a dumb error on my part....

dictoon:
> Remove material alpha mapping
What's the rationale exactly behind this change? It wasn't working properly?

jdent02:
No it wasn't.  Everything in the output file looked correct but either nothing happened or the object disappeared entirely.   Est77 suggested moving it over to a per object option.

PR #169, opened 2018-01-17
Small UI fixes

vallmoy:


properties/world.py
```diff
@@ -42,8 +42,8 @@ def register(cls):
                                                      ("gradient", "Gradient", "Use sky color gradient"),
                                                      ("latlong_map", "Latitude-Longitude Map", "Use latlong map texture"),
                                                      ("mirrorball_map", "Mirror Ball Map", "Use mirror ball texture"),
-                                                     ("sunsky", "Physical Sun/Sky", ""),
-                                                     ("constant_hemisphere", "Per-Hemisphere Constant", "Use constant color per hemisphere")],
+                                                     ("constant_hemisphere", "Per-Hemisphere Constant", "Use constant color per hemisphere"),
```
dictoon: Is this going to break compatibility with existing .blend files, since you're reordering items in the combo?

If so, we need to revert this change since we pledged not to break compatibility again.

vallmoy: Ok, reverted.

PR #176, opened 2018-01-29
Implement OSL shading

jdent02:


operators/__init__.py
```diff
@@ -93,6 +93,22 @@ def execute(self, context):
         return {'FINISHED'}
 
 
+class AppleseedNewOSLNodeTree(bpy.types.Operator):
+
+    bl_idname = "appleseed.add_osl_material_nodetree"
+    bl_label = "Add appleseed OSL Material Node Tree"
+    bl_description = "Create an appleseed osl material node tree and link it to the current material"
```
dictoon: *osl* -> *OSL*

projectwriter.py
```diff
@@ -750,6 +774,61 @@ def __emit_back_material(self, material, material_name, scene, layers, material_
 
         self.__emit_material_element(material_name, bsdf_name, "", "", "", "physical_surface_shader", scene, material, material_node)
 
+    def __emit_osl_shader_group(self, front_material_name, node_list, material_node):
+
```
dictoon: Remove blank line.

projectwriter.py
```diff
@@ -2229,6 +2308,13 @@ def __emit_texture_instance(self, texture, texture_name, bump_bool, node=None, m
         self.__emit_parameter("alpha_mode", "detect")
         self.__close_element("texture_instance")
 
+    def __emit_osl_element(self, front_material_name, surface_name):
+
```
dictoon: Remove blank line.

projectwriter.py
```diff
@@ -2229,6 +2308,13 @@ def __emit_texture_instance(self, texture, texture_name, bump_bool, node=None, m
         self.__emit_parameter("alpha_mode", "detect")
         self.__close_element("texture_instance")
 
+    def __emit_osl_element(self, front_material_name, surface_name):
```
dictoon: `__emit_osl_material` rather?

projectwriter.py
```diff
@@ -2836,6 +2925,12 @@ def __emit_line(self, line):
         self.__emit_indent()
         self._output_file.write(line + "\n")
 
+    def __emit_path(self, path):
```
dictoon: `__emit_search_path`

Move next to `__emit_searchpaths`.

projectwriter.py
```diff
@@ -135,11 +135,19 @@ def __emit_file_header(self):
 
     def __emit_project(self, scene):
         self.__open_element("project")
+        self.__emit_searchpaths()
         self.__emit_scene(scene)
         self.__emit_output(scene)
         self.__emit_configurations(scene)
         self.__close_element("project")
 
+    def __emit_searchpaths(self):
```
dictoon: `__emit_search_paths`

projectwriter.py
```diff
@@ -2878,6 +2973,9 @@ def export_preview(self, scene, file_path, mat, mesh, width, height):
 
         self._textures_set = set()
 
+        osl_reader_path, shader_directories = util.get_osl_searchpaths()
+        appleseed_parent_dir = osl_reader_path.strip("\\bin")
```
dictoon: Isn't there Python functionalities to strip one component from a path? This won't work on Linux or macOS.

projectwriter.py
```diff
@@ -2889,6 +2987,15 @@ def export_preview(self, scene, file_path, mat, mesh, width, height):
                 aspect_ratio = self.__get_frame_aspect_ratio(scene.render)
 
                 self._output_file.write("""<project>
+    <search_paths>
+        <search_path>""")
+                os.path.join(appleseed_parent_dir, 'shaders', 'appleseed')
```
dictoon: How does this work? It looks like you call `os.path.join()` without ever using the result?

properties/nodes/__init__.py
```diff
@@ -28,9 +28,10 @@
 
 import bpy
 import nodeitems_utils
+from nodeitems_utils import NodeItem, NodeCategory
 from bpy.types import NodeTree
 from bpy.app.handlers import persistent
-from ...util import addon_dir, join_names_underscore
+from ...util import addon_dir, join_names_underscore, read_osl_shaders
```
dictoon: Any downside to just doing `import util`?

properties/nodes/__init__.py
```diff
@@ -117,31 +118,57 @@ def poll(cls, context):
 
 # appleseed node categories
```
dictoon: Turn into a docstring.

properties/nodes/__init__.py
```diff
@@ -117,31 +118,57 @@ def poll(cls, context):
...
-        nodeitems_utils.NodeItem("AppleseedBSSRDFNode")]),
-    AppleseedNodeCategory("Volume", "Volume", items=[
-        nodeitems_utils.NodeItem("AppleseedVolumeNode")]),
-    AppleseedNodeCategory("TEXTURES", "Texture", items=[
-        nodeitems_utils.NodeItem("AppleseedTexNode"),
-        nodeitems_utils.NodeItem("AppleseedNormalNode")]),
-    AppleseedNodeCategory("OUTPUTS", "Output", items=[
-        nodeitems_utils.NodeItem("AppleseedMaterialNode")])]
+def node_categories(osl_nodes):
+
```
dictoon: Remove blank line.

properties/nodes/__init__.py
```diff
@@ -117,31 +118,57 @@ def poll(cls, context):
...
+        AppleseedNodeCategory("BSSRDF", "BSSRDF", items=[
+            NodeItem("AppleseedBSSRDFNode")]),
+        AppleseedNodeCategory("Volume", "Volume", items=[
+            NodeItem("AppleseedVolumeNode")]),
+        AppleseedNodeCategory("TEXTURES", "Texture", items=[
+            NodeItem("AppleseedTexNode"),
+            NodeItem("AppleseedNormalNode")]),
+        AppleseedNodeCategory("OUTPUTS", "Output", items=[
+            NodeItem("AppleseedMaterialNode")]),
+        AppleseedNodeCategory("OSL_Shaders", "Osl shaders", items=osl_shaders),
```
dictoon: Osl -> OSL

properties/nodes/__init__.py
```diff
@@ -192,12 +219,14 @@ def appleseed_scene_loaded(dummy):
 from . import normal
 from . import volume
 from . import material
+from . import oslnode
 
+osl_node_names = []
 
 def register():
+    node_list = read_osl_shaders()
```
dictoon: Move closer to where it's needed.

properties/nodes/__init__.py
```diff
@@ -218,6 +247,11 @@ def register():
     normal.register()
     volume.register()
     material.register()
+    if node_list:
```
dictoon: What's the purpose of this test?

properties/nodes/oslnode.py
```diff
@@ -0,0 +1,445 @@
+
+#
+# This source file is part of appleseed.
+# Visit http://appleseedhq.net/ for additional information and resources.
+#
+# This software is released under the MIT license.
+#
+# Copyright (c) 2014-2017 The appleseedhq Organization
```
dictoon: Fix date: 2018

util.py
```diff
@@ -35,6 +35,159 @@
...
+    if bpy.context.user_preferences.addons['blenderseed'].preferences.appleseed_binary_directory == "":
+        print("appleseed binary path not set.  Rendering and OSL features will not be avaialable")
+        return
+            
+    osl_path, shader_directories = get_osl_searchpaths()
+    nodes = []
+    osl_files = []
+
+    for shader_dir in shader_directories:
+            for file in os.listdir(shader_dir):
```
dictoon: Fix indentation.

util.py
```diff
@@ -35,6 +35,159 @@
...
+                if file.endswith(".osl"):
+                    filename = os.path.join(shader_dir, file)
+                    print(filename)
+                    osl_files.append((shader_dir, filename))
+
+    for file in osl_files:
+            cmd = ('%s\\oslc %s' % (osl_path, file[1]))
```
dictoon: Fix indentation.

util.py
```diff
@@ -35,6 +35,159 @@
...
+    appleseed_parent_dir = osl_path.strip("\\bin")
+    shader_directories = (os.path.join(appleseed_parent_dir, 'shaders', 'appleseed'), os.path.join(appleseed_parent_dir, 'shaders', 'blenderseed'))
+
+    return osl_path, shader_directories
+
+def read_osl_shaders():
+    """Compiles any .osl files and reads all .oso parameters"""
+
+    if bpy.context.user_preferences.addons['blenderseed'].preferences.appleseed_binary_directory == "":
+        print("appleseed binary path not set.  Rendering and OSL features will not be avaialable")
```
dictoon: *available*

util.py
dictoon: Use `os.path.join()` to join path fragments in a portable manner. This won't work on Linux or macOS.

util.py
```diff
@@ -35,6 +35,159 @@
...
+            working_dir = file[0]
+            process = subprocess.Popen(cmd, cwd=working_dir, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+            print("Compiled {0} shader".format(file[1]))
+
+    for shader_dir in shader_directories:
+        for file in os.listdir(shader_dir):
+            if file.endswith(".oso"):
+                filename = os.path.join(shader_dir, file)
+                content = []
+                cmd = ('%s\\oslinfo -v %s' % (osl_path, filename))
```
dictoon: Use `os.path.join()`.

util.py
```diff
@@ -35,6 +35,159 @@
...
+                process = subprocess.Popen(cmd, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+                while 1:
```
dictoon: `while True:`

util.py
```diff
@@ -35,6 +35,159 @@
...
+                    content.append(line)
+                    if not line: 
+                        break
+                shader_params = create_osl_dict(filename, content)
+                nodes.append(shader_params)
+
+    return nodes
+
+def create_osl_dict(filename, content=None):
+
```
dictoon: Remove redundant blank line.

util.py
```diff
@@ -35,6 +35,159 @@
...
+# ------------------------------------
+
+def get_osl_searchpaths():
+    osl_path = bpy.context.user_preferences.addons['blenderseed'].preferences.appleseed_binary_directory
+    appleseed_parent_dir = osl_path.strip("\\bin")
+    shader_directories = (os.path.join(appleseed_parent_dir, 'shaders', 'appleseed'), os.path.join(appleseed_parent_dir, 'shaders', 'blenderseed'))
+
+    return osl_path, shader_directories
+
+def read_osl_shaders():
```
dictoon: Instrument this function to make it clear what folders it's looking into, what files it's processing and what nodes were discovered/compiled/analyzed.

properties/nodes/__init__.py
jdent02: To prevent an exception if the following 'for' loop tries to iterate over an empty list.

util.py
jdent02: Are you meaning add print() lines to show the progress of each analysis in the console?

projectwriter.py
jdent02: Can't.  already have a function higher up with that name.  I purposely modeled the OSL element output after the built in shader output

jdent02:
Should have all your comments covered.

projectwriter.py
```diff
@@ -135,19 +135,25 @@ def __emit_file_header(self):
...
+    def __emit_search_paths(self):
+        osl_path, shader_directories = util.get_osl_search_paths()
         self.__open_element("search_paths")
         for path in shader_directories:
-            self.__emit_path(path)
+            self.__emit_search_path(path)
         self.__close_element("search_paths")
 
+    def __emit_search_path(self, path):
+        self.__emit_line("<search_path>")
```
dictoon: You should use `__open_element()` instead of `__emit_line()` + `__indent()`.

properties/nodes/__init__.py
dictoon: Iterating over an empty list is perfectly valid. Iterating over a `None` object isn't, but then I would expect `read_osl_shaders()` to always return a valid (but possibly empty) list.

properties/nodes/oslnode.py
```diff
@@ -5,7 +5,7 @@
 #
 # This software is released under the MIT license.
 #
-# Copyright (c) 2014-2017 The appleseedhq Organization
+# Copyright (c) 2014-2018 The appleseedhq Organization
```
dictoon: You can remove the `2014-` part.

Sorry, I should simply have pointed to our guidelines for copyright notices:
https://github.com/appleseedhq/appleseed/wiki/Copyright-Notices

util.py
```diff
@@ -40,63 +40,67 @@
...
-
 # ------------------------------------
 # OSL shader reader
 # ------------------------------------
 
-def get_osl_searchpaths():
+def get_osl_search_paths():
     osl_path = bpy.context.user_preferences.addons['blenderseed'].preferences.appleseed_binary_directory
-    appleseed_parent_dir = osl_path.strip("\\bin")
+    appleseed_parent_dir = os.path.dirname(os.path.dirname(osl_path))
```
dictoon: That's probably where the error I talked about on Slack is. In fact there seems to be two problems:

1. `osl_path` is simply `preferences.appleseed_binary_directory`. That doesn't seem correct.
2. The algorithm for finding appleseed's root directory needs to account for two cases: "retail" versions of appleseed, and "development" versions. The exact algorithm for determining appleseed's root directory is implemented [here](https://github.com/appleseedhq/appleseed/blob/af3c49ae8bb08f3bc50b23fa05fc564d40561758/src/appleseed.shared/application/application.cpp#L134-L156). Basically, you strip directories until you read `bin`, then you strip once more.

util.py
```diff
@@ -40,63 +40,67 @@
...
-                    print(filename)
-                    osl_files.append((shader_dir, filename))
-
-    for file in osl_files:
-            cmd = ('%s\\oslc %s' % (osl_path, file[1]))
-            print(cmd)
-            working_dir = file[0]
-            process = subprocess.Popen(cmd, cwd=working_dir, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-            print("Compiled {0} shader".format(file[1]))
+        for file in os.listdir(shader_dir):
```
dictoon: I would suggest skipping directories that don't exist, to avoid the error I pasted in Slack.

util.py
```diff
@@ -40,63 +40,67 @@
...
+            if file.endswith(".osl"):
+                print("appleseed Compiling {0}".format(file))
+                filename = os.path.join(shader_dir, file)
+                oslc_path = os.path.join(osl_path, "oslc")
+                cmd = (oslc_path, filename)
+                process = subprocess.Popen(cmd, cwd=shader_dir, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
dictoon: Eventually we'll need to wrap the `subprocess.Popen()` call in a try/catch block.

util.py
```diff
@@ -40,63 +40,67 @@
...
             if file.endswith(".oso"):
+                print("appleseed Reading {0}".format(file))
                 filename = os.path.join(shader_dir, file)
                 content = []
-                cmd = ('%s\\oslinfo -v %s' % (osl_path, filename))
-                process = subprocess.Popen(cmd, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-                while 1:
+                oslinfo_path = os.path.join(osl_path, "oslinfo")
+                cmd = (oslinfo_path, '-v', filename)
+                process = subprocess.Popen(cmd, cwd=osl_path, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
dictoon: Similarly here, we'll need to wrap this in a try/catch block.

util.py
```diff
@@ -40,63 +40,67 @@
...
+                while True:
                     line = process.stdout.readline()
                     line = line.strip()
                     content.append(line)
-                    if not line: 
+                    if not line:
```
dictoon: The logic is slightly suspicious here. You check if `line` is empty after having appended it to `content`, why?

util.py
```diff
@@ -40,63 +40,67 @@
...
                         break
-                shader_params = create_osl_dict(filename, content)
-                nodes.append(shader_params)
+                shader_params = create_osl_dict(file, content)
+                if shader_params != {}:
```
dictoon: `if shader_params:` should be enough.

util.py
```diff
@@ -171,22 +175,23 @@ def create_osl_dict(filename, content=None):
 
     return d
 
+
 def reverseValidate(pname):
     if pname == "inMin":
         return "min"
     if pname == "inMax":
-        return "max";
+        return "max"
```
dictoon: Well spotted :)

util.py
jdent02: Agree on 2.  For 1 it is correct as all osl related apps are in the bin directory.  or is that different for your dev versions?

dictoon: > Agree on 2. For 1 it is correct as all osl related apps are in the bin directory. or is that different for your dev versions?

Oh ok, I misunderstood what `osl_path` was then. I suggest `appleseed_bin_dir` instead of `osl_path`, it'll be a lot clearer.

util.py
jdent02: I moved this to in front of the .strip option.  For whatever reason if this line isn't in there the oslinfo process drops into an endless loop.

util.py
```diff
@@ -59,42 +63,48 @@ def read_osl_shaders():
...
+                    filename = os.path.join(shader_dir, file)
+                    oslc_path = os.path.join(appleseed_bin_dir, "oslc")
+                    cmd = (oslc_path, filename)
+                    try:
+                        process = subprocess.Popen(cmd, cwd=shader_dir, bufsize=1, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+                        print("appleseed Compiled {0}".format(file))
+                    except:
+                        print("ERROR: Failed to compile {0}".format(file))
+
+    print("appleseed Parsin OSL shaders")
```
dictoon: *Parsing*

PR #179, opened 2018-01-30
Fix for texture used as color and bump map

jdent02:


projectwriter.py
```diff
@@ -2209,7 +2209,7 @@ def __emit_texture(self, texture, bump_bool, scene, node=None, material_name=Non
                 color_space = 'srgb'
 
             filepath = util.realpath(texture.image.filepath)
-            texture_name = texture.name + "_bump" if bump_bool is True else texture.name
+            texture_name = texture.name
```
dictoon: Do we need the `texture_name` variable then?

jdent02:
Okay to merge this?

dictoon:
Absolutely! Merging now.

dictoon:
Thanks for the quick fix.

PR #183, opened 2018-02-04
OSL updates

jdent02:
Minor unregistering addon fix

### DIFF
--- a/projectwriter.py
+++ b/projectwriter.py
@@ -688,18 +688,18 @@ class Writer(object):
         else:
             return asr_mat.use_light_emission and scene.appleseed.export_emitting_obj_as_lights
 
-    def __is_node_material(self, asr_mat):
+    def __is_node_material(self, asr_mat, node_tree):
         if asr_mat.node_tree != "":
-            for node in bpy.data.node_groups[asr_mat.node_tree].nodes:
+            for node in bpy.data.node_groups[node_tree].nodes:
                 if node.node_type == "osl_surface":
                     return True
                 elif node.node_type == 'material':
                     return True
         return False
 
-    def __is_osl_material(self, asr_mat):
-        if asr_mat.node_tree != "":
-            for node in bpy.data.node_groups[asr_mat.node_tree].nodes:
+    def __is_osl_material(self, asr_mat, node_tree):
+        if asr_mat.osl_node_tree != "":
+            for node in bpy.data.node_groups[node_tree].nodes:
                 if node.node_type == "osl_surface":
                     return True
         return False
@@ -720,9 +720,9 @@ class Writer(object):
         """Write the material."""
 
         asr_mat = material.appleseed
-        asr_node_tree = asr_mat.node_tree
-        use_nodes = self.__is_node_material(asr_mat)
-        use_osl = self.__is_osl_material(asr_mat)
+        asr_node_tree = asr_mat.node_tree if not asr_mat.use_osl else asr_mat.osl_node_tree
+        use_nodes = self.__is_node_material(asr_mat, asr_node_tree)
+        use_osl = self.__is_osl_material(asr_mat, asr_node_tree)
         layers = asr_mat.layers
 
         material_node = None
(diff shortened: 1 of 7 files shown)


projectwriter.py
```diff
@@ -2457,7 +2457,7 @@ def __emit_camera(self, scene):
             else:
                 focal_distance = camera.data.dof_distance
 
-        self.__open_element('camera name="' + camera.name + '" model="{}_camera"'.format(cam_model))
+        self.__open_element('camera name="' + camera.name + '" model="{0}_camera"'.format(cam_model))
```
dictoon: You could also pass the camera name to `format()` here.

util.py
```diff
@@ -67,21 +67,32 @@ def read_osl_shaders():
...
+                        while True:
+                            line = process.stdout.readline()
+                            if not line:
+                                break
+                            line = line.strip().decode("utf-8")
+                            if "error" in line:
+                                print(line)
+                                print("ERROR: Failed to compile {0}".format(file))
+                                continue
                         print("appleseed Compiled {0}".format(file))
```
dictoon: Unfortunately this message will still be printed in case of errors.

util.py
```diff
@@ -67,21 +67,32 @@ def read_osl_shaders():
...
                     except:
                         print("ERROR: Failed to compile {0}".format(file))
```
dictoon: It would make sense here to print the exception message as well.

util.py
```diff
@@ -107,6 +107,11 @@ def read_osl_shaders():
                             if not line:
                                 break
                             line = line.strip()
+                            line = line.strip().decode("utf-8")
```
dictoon: Calling `strip()` twice on `line`.

projectwriter.py
```diff
@@ -689,18 +689,9 @@ def __is_light_emitting_material(self, asr_mat, scene, material_node=None):
             return asr_mat.use_light_emission and scene.appleseed.export_emitting_obj_as_lights
 
     def __is_node_material(self, asr_mat, node_tree):
-        if asr_mat.node_tree != "":
+        if asr_mat.node_tree != "" or asr_mat.osl_node_tree != "":
```
dictoon: Should `__is_node_material()` be called if neither `asr_mat.node_tree` nor `asr_mat.osl_node_tree` is set?

jdent02: I suppose not but I didn’t change that part. Is_node_material has always been called during any render

util.py
jdent02: Good point

projectwriter.py
jdent02: I don't think there's much need to change this behavior.  If either node tree has an entry it has to run the check even if the node tree isn't 'active', for instance the 'use OSL' button is on but the node tree  uses built in items.

jdent02:
Fixes in.

util.py
```diff
@@ -84,10 +84,10 @@ def read_osl_shaders():
                             if "error" in line:
                                 print(line)
                                 print("ERROR: Failed to compile {0}".format(file))
-                                continue
+                                break
```
dictoon: That only works if error messages are on a single line.

util.py
```diff
@@ -69,21 +67,32 @@ def read_osl_shaders():
...
+                            line = line.strip().decode("utf-8")
+                            if "error" in line:
+                                print(line)
+                                print("ERROR: Failed to compile {0}".format(file))
+                                break
                         print("appleseed Compiled {0}".format(file))
                     except:
-                        print("ERROR: Failed to compile {0}".format(file))
+                        print("ERROR: Failed to compile {0}.  Oslc did not launch".format(file))
```
dictoon: One extra space after ".".

It would also be nice to include the exception details (in the oslinfo case too).

util.py
jdent02: Every compile error I've seen so far has been.

util.py
jdent02: Isn't that covered by the error line print already?  Or are you suggesting to store the error lines and then roll them into this line instead of individually printing?

dictoon: No, you're mixing two different issues:
1. Compilation error messages emitted by `oslc` to the process' `stdout`: you properly handle those
2. Other errors, in the form of exceptions, including (but not only!) failures to start `oslc`: in that case you print a generic failure message but that's not enough, it would be nice to also have the exception's message.

jdent02: So I tested an attempt to launch oslc with the .exe moved so it couldn't be found  It didn't generate an error, just a few lines on how the command didn't point to a recognized program.  The subprocess technically still launched so the exception (line 89) didn't trigger.

dictoon: Some cases that will lead to exceptions: https://docs.python.org/2/library/subprocess.html#exceptions

jdent02:
Not sure if the stuff I added will work.  I found it online but I don't quite understand how it works

dictoon:
Thanks, let's try with that and refine as necessary!

PR #185, opened 2018-02-12
Remove material layer system

jdent02:


util.py
```diff
@@ -112,7 +112,7 @@ def read_osl_shaders():
                             if not line:
                                 break
                             line = line.strip().decode("utf-8")
-                            if "error" in line:
+                            if "ERROR" in line:
```
dictoon: I feel that deserves a separate commit.

properties/nodes/__init__.py
```diff
@@ -296,9 +295,12 @@ def register():
...
     node_list = read_osl_shaders()
-    for node in node_list:
-        node_name, node_category = oslnode.generate_node(node)
-        osl_node_names.append([node_name, node_category])
+    if node_list:
+        for node in node_list:
+            node_name, node_category = oslnode.generate_node(node)
+            osl_node_names.append([node_name, node_category])
+    else:
+        print("ERROR: OSL Nodes list is empty")
```
dictoon: *Nodes* -> *nodes*

PR #187, opened 2018-02-23
Remove built in material nodes

jdent02:
Remove built in nodes from node systems

Remove built in nodes from ui and exporter


jdent02:
#186

projectwriter.py
```diff
@@ -681,17 +681,14 @@ def __emit_visibility_flags(self, object):
...
     """Material(s) writing"""
 
-    def __is_light_emitting_material(self, asr_mat, scene, material_node=None):
-        if material_node is not None:
-            if material_node.inputs["Emission Strength"].socket_value > 0.0 or material_node.inputs["Emission Strength"].is_linked:
-                return scene.appleseed.export_emitting_obj_as_lights
-        else:
-            return asr_mat.use_light_emission and scene.appleseed.export_emitting_obj_as_lights
+    def __is_light_emitting_material(self, asr_mat, scene):
+
```
dictoon: Remove blank line.

It might be a good time to run autopep8 on the code.

projectwriter.py
```diff
@@ -846,307 +836,213 @@ def __emit_front_material_tree(self, material, material_name, scene, material_no
...
+            # Volume
+            if asr_mat.volume_phase_function_model != 'none':
+                volume_name = "{0}_volume".format(material_name)
+                self.__emit_volume(material, volume_name, scene)
+
+            return bsdf_name, bssrdf_name, volume_name
+
+    def __emit_back_material_bsdf_tree(self, material, material_name, scene):
         """Emits mateiral tree for the back side of the object."""
```
dictoon: Typo: *material*

properties/materials.py
```diff
@@ -30,1512 +30,1503 @@
 
 
 def refresh_preview(self, context):
-    """
-    Triggers a refresh of the material preview.
-    """
+  """
```
dictoon: Fix indentation in whole file (should be 4 spaces).

jdent02:
I'm having some issues with my PEP8 formatter.  It keeps splitting lines when it shouldn't even though length checking is supposed to be off.

For example: default=(0.8, 0.8, 0.8) gets split to

default=(
        0.8, 0.8, 0.8)

dictoon:
Hm, that's annoying. I use `--max-line-length 150` here but I guess you're already doing something similar.

jdent02:
Should be good now.

dictoon:
Cool!

In the future, could you _add_ commits to address review comments, and only rebase/squash just before merging? That would make reviews easier.

Merging now, thanks for the great work!

PR #191, opened 2018-03-02
Fixes

jdent02:


util.py
```diff
@@ -74,7 +74,7 @@ def read_osl_shaders():
                     print("appleseed Compiling {0}".format(file))
                     filename = os.path.join(shader_dir, file)
                     oslc_path = os.path.join(appleseed_bin_dir, "oslc")
-                    cmd = (oslc_path, filename)
+                    cmd = (oslc_path, "{0}".format(filename))
```
dictoon: Is this really needed? It looks suspicious.

util.py
```diff
@@ -104,7 +104,7 @@ def read_osl_shaders():
                     filename = os.path.join(shader_dir, file)
                     content = []
                     oslinfo_path = os.path.join(appleseed_bin_dir, "oslinfo")
-                    cmd = (oslinfo_path, '-v', filename)
+                    cmd = (oslinfo_path, '-v', "{0}".format(filename))
```
dictoon: Same here.

properties/nodes/__init__.py
```diff
@@ -198,8 +198,9 @@ def register():
     node_list = read_osl_shaders()
     if node_list:
         for node in node_list:
-            node_name, node_category = oslnode.generate_node(node)
-            osl_node_names.append([node_name, node_category])
+            if node != {}:
```
dictoon: Isn't

    if node:

enough?

util.py
jdent02: Was a way of handling spaces in filepaths on Windows

util.py
jdent02: Same here.

properties/nodes/__init__.py
jdent02: You're right.

util.py
dictoon: I'm not convinced. If `filename` is a string then `"{0}".format(filename)` has exactly the same value as `filename` (unless I'm missing something obvious).

jdent02: Okay I modified it.  Now it works correctly.

dictoon: You modified it below (makes a lot more sense now) but not here.

jdent02: Whoops....Fixed.

dictoon:
Can you fold the fixes commit into another one and then merge this PR please?

Issue #154, opened 2017-12-30
Render tile buckets are saved with cancelled render


vallmoy:
![render tile error](https://user-images.githubusercontent.com/30292603/34454252-b1a321dc-ed67-11e7-8276-6a9a59a5392f.png)

jdent02:
Update: Removing tile highlights works on full image renders.  Border renders however do not.

jdent02:
FYI completion of this is waiting on some modifications to appleseed itself.

jdent02:
Decision was made to remove tile highlighting from the in progress render.

jdent02:
#238

PR #199, opened 2018-03-18
Remove tile highlights when canceling render-DO NOT MERGE

jdent02:
#154

### DIFF
--- a/render.py
+++ b/render.py
@@ -34,6 +34,7 @@ import shutil
 import tempfile
 import threading
 import time
+import numpy as np
 from math import ceil
 from shutil import copyfile
 
@@ -206,9 +207,7 @@ class RenderAppleseed(bpy.types.RenderEngine):
         self.total_pixels = (max_x - min_x + 1) * (max_y - min_y + 1) * self.total_passes
 
         # Create secondary image buffer
-        self.image_buffer = []
-        for x in range(0, (int(width * height) * 4)):
-            self.image_buffer.append(0.0)
+        self.image_buffer = np.zeros(shape=[4, max_y - min_y + 1, max_x - min_x + 1], dtype=np.float32)
 
         # Launch appleseed.cli.
         threads = 'auto' if scene.appleseed.threads_auto else str(scene.appleseed.threads)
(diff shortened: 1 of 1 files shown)


render.py
```diff
@@ -34,6 +34,7 @@
 import tempfile
 import threading
 import time
+import numpy as np
```
dictoon: Order alphabetically.

render.py
```diff
@@ -302,14 +324,20 @@ def __process_tile_data_chunk(self, process, min_x, min_y, max_x, max_y):
         floats = array.array('f')
         floats.fromstring(tile_data)
         pix = []
+        y_start = y0
```
dictoon: I think `x_start` and `y_start` aren't good names.

Isn't `y_start` equal to `y0 + y`? And `x_start` could simply be called `x`.

render.py
```diff
@@ -333,6 +361,7 @@ def __process_tile_data_chunk(self, process, min_x, min_y, max_x, max_y):
         return True
 
     def __process_tile_highlight_chunk(self, process, min_x, min_y, max_x, max_y):
+
```
dictoon: Remove blank line (as per PEP8).

render.py
```diff
@@ -248,6 +252,20 @@ def __render_project_file(self, scene, project_filepath, project_dir=None):
                 os.read(process.stdout.fileno(), chunk_size)
                 continue
 
+        if self.test_break():
+            display_buffer = []
```
dictoon: Maybe you can get rid of `display_buffer` and work directly with `layer.rect`?

render.py
jdent02: Nope.  In Blender y starts at the bottom. If I did y+y0 it would start from the top left corner.  That ‘y’ variable is for pulling the correct pixels out of the array that’s made from the appleseed output and translating that to Blenders pixel order

render.py
jdent02: Worth a try

jdent02: It doesn't work.  Trying to append directly into layer.rect gives me AttributeError: 'bpy_prop_array' object has no attribute 'append'.  I'm trying to find more info on this item

jdent02: There's no documentation on that structure.  Even in the official Blender Python docs

dictoon: Ah ok, then it's good like that I guess.

PR #206, opened 2018-03-27
Expose render stamp and denoiser features

jdent02:
#198 
#200


docs/panels/render/render.rst
```diff
@@ -0,0 +1,15 @@
+Render
+======
+
+.. image:: /_static/screenshots/render_panels/render.JPG
+
+|
+
+- Export appleseed Files Only
+	- This options tells blenderseed to create the scene files for rendering and export the geometry only.  The rendering process will not be triggered.  This allows you to render the scene at a later time using the appleseed command line or appleseed Studio.
```
dictoon: *appleseed.studio*

docs/panels/object/appleseed_motion_blur.rst
```diff
@@ -0,0 +1,2 @@
+Appleseed Motion Blur
```
dictoon: If possible, appleseed should be spelled in lower case, even when it's a title's first word.

docs/panels/render/render.rst
```diff
@@ -0,0 +1,15 @@
...
+- Render
+	- This button triggers the exporting process, and then the actual render if the export only button is not selected.
+- Animation
+	- Render a sequential output of frames (determined by the start and end frame settings).
+- Display
+	- This determines how the blender desktop will change to accomodate the rendered image.
```
dictoon: *Blender*

properties/scene.py
```diff
@@ -51,368 +51,409 @@ def camera_enumerator(self, context):
 
 class AppleseedRenderSettings(bpy.types.PropertyGroup):
 
-    @classmethod
-    def register(cls):
-        bpy.types.Scene.appleseed = bpy.props.PointerProperty(name="appleseed_render_settings",
-                                                              description="appleseed Render Settings",
-                                                              type=cls)
+    def update_stamp(self, context):
+        self.render_stamp = "{0}|{1}".format(self.render_stamp, self.render_stamp_patterns)
```
dictoon: I'm not sure to understand this code. What does an example render stamp look like?

render.py
```diff
@@ -183,6 +183,9 @@ def __render_project_file(self, scene, project_filepath, project_dir=None):
         # Total tiles.
         self.rendered_tiles = 0
 
+        # Is the denoiser on
+        self.do_denoise = scene.appleseed.enable_denoiser
```
dictoon: Why not use `scene.appleseed.enable_denoiser` directly instead of introducing an alias?

ui/render.py
```diff
@@ -1,4 +1,3 @@
-
 #
```
dictoon: All files should start with a blank line.

ui/render.py
```diff
@@ -98,6 +97,107 @@ def draw(self, context):
...
+
+        col = split.column()
+        col.prop(rd, "use_file_extension")
+        col.prop(rd, "use_render_cache")
+
+        layout.template_image_settings(image_settings, color_management=False)
+        if rd.use_multiview:
+            layout.template_image_views(image_settings)
+
+        if file_format == 'QUICKTIME':
```
dictoon: I find it weird that we have code related to audio codecs in a file called `render.py` in a renderer plugin :)

docs/about.rst
```diff
@@ -3,7 +3,7 @@ About blenderseed
 
 The orginal appleseed integration into blender was written by François Beaune when appleseed was in an alpha state. It was a primitive plugin, matching an equally primitive (at the time) renderer.  Over the next several years appleseed grew and improved in leaps and bounds, and blenderseed has grown along with it thanks to the dedication and efforts of many volunteer authors:
 
-	- Projec Lead: vallmoy(Luke Kliber)
+	- Project Lead: vallmoy(Luke Kliber)
```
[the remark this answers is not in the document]
jdent02: Sure

properties/scene.py
jdent02: {render-time}|{lib-version}.  Can be anything really, this function just automatically inserts the variables so the user doesn’t have to remember the formatting themselves

ui/render.py
jdent02: Its to make sure it retains the same functionality as Blenders built in output tab. Although the reason I added the panel here has been removed so I suppose I can do the same here

properties/scene.py
dictoon: The problem is that placeholders such as `{render-time}` are just values, they don't contain any label in front. So what we want in a render stamp is a string with placeholders interspersed at the right places, e.g.

    "appleseed {lib-version} | Time: {render-time}"

Also note spaces around `|`.

jdent02: I figured the rest of the text could be added by the user depending on how they wanted it to appear

jdent02: Yep.  Selecting an option in the dropdown just inserts it into the string.  The string itself appears in the UI and is modifiable by the user.

jdent02: So they can use multiple variables in the string if they want.  Like I said I thought it would be more convenient for the user if they didn't have to remember the variables

dictoon: > Selecting an option in the dropdown just inserts it into the string. The string itself appears in the UI and is modifiable by the user.

Ok, that's perfect.

> I thought it would be more convenient for the user if they didn't have to remember the variables

Yes, this is obviously a good idea.

The only thing I'm not convinced about is this:

    "{0}|{1}".format(self.render_stamp, self.render_stamp_patterns)

Is that some kind of default render stamp? It just won't look good, that's why I'm concerned.

PR #207, opened 2018-03-29
Fix registration of project writer

oktomus:
`project_file_writer` doesn't exist anymore.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -48,7 +48,6 @@ if "bpy" in locals():
     imp.reload(render)
     imp.reload(util)
     imp.reload(preferences)
-    imp.reload(projectwriter)
 
 else:
     import bpy
--- a/__init__.py
+++ b/__init__.py
@@ -48,7 +48,7 @@ if "bpy" in locals():
     imp.reload(render)
     imp.reload(util)
     imp.reload(preferences)
-    imp.reload(project_file_writer)
+    imp.reload(projectwriter)
 
 else:
     import bpy


__init__.py
```diff
@@ -48,7 +48,7 @@
     imp.reload(render)
     imp.reload(util)
     imp.reload(preferences)
-    imp.reload(project_file_writer)
+    imp.reload(projectwriter)
```
dictoon: Hm, good catch. What was the impact of this error?

oktomus: Blenderseed not registering and not working ^^ But actually, we don't use it so we can remove it.

dictoon: I mean, blenderseed has been working fine so far no?

oktomus: If bpy is not in your sys, yes. But for me, I am using blender 2.79 and I guess bpy is by default in my path.

dictoon: I'm using Blender 2.79 too and I've never seen the problem.

Is the problem you're mentioning related to loading the plugin at startup, loading the plugin when installing it, or reloading the plugin when the code has changed?

oktomus: Loading the plugin at startup. In any case, project_file_writer doesn't exists. Maybe you have a __pycache__ somewhere saving you from this.

oktomus: Reloading sorry, not at startup !

oktomus:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/share/blender/2.79/scripts/modules/bpy/utils/__init__.py", line 266, in load_scripts
    _addon_utils.reset_all(reload_scripts=reload_scripts)
  File "/usr/share/blender/2.79/scripts/modules/addon_utils.py", line 437, in reset_all
    importlib.reload(mod)
  File "/usr/lib/python3.6/importlib/__init__.py", line 166, in reload
    _bootstrap._exec(spec, module)
  File "<frozen importlib._bootstrap>", line 618, in _exec
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/oktomus/.config/blender/2.79/scripts/addons/blenderseed/__init__.py", line 51, in <module>
    imp.reload(project_file_writer)
NameError: name 'project_file_writer' is not defined

location: <unknown location>:-1
```

dictoon: > Reloading sorry, not at startup !

Ah that makes more sense! Ok, merging this, thank you. Does reloading work now?

oktomus: So if it's at reloading, I should not delete the line. I will remove my last commit.

PR #208, opened 2018-03-29
Implement protocol v2 decoding and reading

oktomus:
Needs to be merged along with [ Allow sending AOV tiles to stdout #1906 ](https://github.com/appleseedhq/appleseed/pull/1906)

### DIFF
--- a/render.py
+++ b/render.py
@@ -240,19 +240,19 @@ class RenderAppleseed(bpy.types.RenderEngine):
                 # Tile data, protocol v1
                 if not self.__process_tile_data_chunk(process, min_x, min_y, max_x, max_y):
                     break
-            elif (chunk_type == 2) or (chunk_type == 10):
+            elif chunk_type == 2 or chunk_type == 10:
                 # Tile highlight, protocol v1 and v2
                 if not self.__process_tile_highlight_chunk(process, min_x, min_y, max_x, max_y):
                     break
-            elif (chunk_type == 11):
+            elif chunk_type == 11:
                 # Tiles header, protocol v2
                 if not self.__process_tiles_header(process):
                     break
-            elif (chunk_type == 12):
+            elif chunk_type == 12:
                 # Plane definition header, protocol v2
                 if not self.__process_plane_header(process):
                     break
-            elif (chunk_type == 13):
+            elif chunk_type == 13:
                 # Plane pixels, protocol v2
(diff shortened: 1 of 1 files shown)


render.py
```diff
@@ -235,14 +235,27 @@ def __render_project_file(self, scene, project_filepath, project_dir=None):
...
 
+            # Protocol V1
             if chunk_type == 1:
-                # Tile data (protocol v1).
+                # Tile data, protocol v1
                 if not self.__process_tile_data_chunk(process, min_x, min_y, max_x, max_y):
                     break
-            elif chunk_type == 2:
-                # Tile highlight (protocol v1).
+            elif (chunk_type == 2) or (chunk_type == 10):
```
dictoon: Remove redundant parentheses here and in other relevant places.

render.py
```diff
@@ -387,6 +402,119 @@ def __process_tile_highlight_chunk(self, process, min_x, min_y, max_x, max_y):
...
+
+        aov_name = os.read(process.stdout.fileno(), aov_name_len)
+
+        return True
+
+    def __process_plane_data_chunk(self, process, min_x, min_y, max_x, max_y):
+        # Read and decode plane header and content.
+        # Protocol v2
+        # Contains:
+        #  Aov index
```
dictoon: *AOV*

render.py
```diff
@@ -387,6 +402,119 @@ def __process_tile_highlight_chunk(self, process, min_x, min_y, max_x, max_y):
 
         return True
 
+    def __process_tiles_header(self, process):
+        # Read and decode tiles header
```
dictoon: In Python it's actually better to use docstrings: https://www.python.org/dev/peps/pep-0257/

render.py
```diff
@@ -387,6 +402,119 @@ def __process_tile_highlight_chunk(self, process, min_x, min_y, max_x, max_y):
...
+        #  Tile pixels
+        tile_header = struct.unpack("IIIIII", os.read(process.stdout.fileno(), 6 * 4))
+        tile_aov_index = tile_header[0]
+        tile_x = tile_header[1]
+        tile_y = tile_header[2]
+        tile_w = tile_header[3]
+        tile_h = tile_header[4]
+        tile_c = tile_header[5]
+
+        # We process only the beauty AOV
```
dictoon: End sentence with a point.

oktomus:
@dictoon Yes, this one can be merged before the other, but not the opposite.

PR #217, opened 2018-04-10
Add texture converter and AOV display

jdent02:


jdent02:
#215

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
+        scene = context.scene
+        textures = scene.appleseed
+        
+        tool_dir, shader_dir = get_osl_search_paths()
+
+        for tex in textures.textures:
+            filename = bpy.path.abspath(tex.name)
+            cmd = ['maketx', '--oiio --monochrome-detect -u --constant-color-detect --opaque-detect', '"{0}"'.format(filename)]
+            if tex.input_space != 'linear':
+                cmd.insert(-1, '--colorconvert {0} linear --unpremult'.format(tex.input_space))
```
dictoon: These `insert()` calls are a bit suspicious: passing -1 for the first argument doesn't seem correct, it should be 0 if you want to insert items at the front.

Also, why not just append arguments? Why insert them at the front of the array?

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
+        collection = scene.appleseed.textures
+
+        existing_textures = [x.name for x in collection]
+
+        for tree in bpy.data.node_groups:
+            for node in tree.nodes:
+                for param in node.parameter_types:
+                    if node.parameter_types[param] == 'string':
+                        string = getattr(node, param)
+                        if string.endswith(('.jpg', '.png', '.tif')):
```
dictoon: Why only these three extensions? What about bmp, exr, hdr, etc.?

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
+                            if string not in existing_textures:
+                                collection.add()
```
dictoon: That looks a bit flimsy :) Isn't there a nicer way to append an object with a `name` property?

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
+            process = subprocess.Popen(" ".join(cmd), cwd=tool_dir, shell=True, bufsize=1)
+            process.wait()
+
+        return {'FINISHED'}
 
 
-class AppleseedAddMatLayer(bpy.types.Operator):
+class AppleseedRefreshTexture(bpy.types.Operator):
     """
     Operator for adding material layers.
```
dictoon: Comment looks stale.

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
+                                collection.add()
+                                num = collection.__len__()
+                                collection[num - 1].name = string
+
+        return {'FINISHED'}
+
+
+class AppleseedAddTexture(bpy.types.Operator):
+    """
+    Operator for adding material layers.
```
dictoon: Comment looks stale.

operators/__init__.py
```diff
@@ -27,49 +27,107 @@
...
-        collection[num - 1].bsdf_name = "BSDF Layer " + str(num)
+        collection[num - 1].name = ""
 
         return {'FINISHED'}
 
 
-class AppleseedRemoveMatLayer(bpy.types.Operator):
+class AppleseedRemoveTexture(bpy.types.Operator):
     """
     Operator for removing material layers.
```
dictoon: Comment looks stale.

projectwriter.py
```diff
@@ -806,6 +806,11 @@ def __get_osl_node_params(self, node):
                 parameter = getattr(node, item)
                 if item in node.filepaths:
                     parameter = bpy.path.abspath(parameter)
+                    if scene.appleseed.sub_textures is True:
+                        if parameter.endswith(('.jpg', '.png', '.tif')):
```
dictoon: I'd suggesting factorizing the list of supported texture extensions to avoid having to update it in multiple places.

Also, are there redundant parentheses here?

projectwriter.py
```diff
@@ -2177,8 +2182,31 @@ def __emit_frame_element(self, scene):
         if scene.render.use_border:
             min_x, min_y, max_x, max_y = self.__get_border_limits(scene, width, height)
             self.__emit_parameter("crop_window", "{0} {1} {2} {3}".format(min_x, min_y, max_x, max_y))
+        if scene.appleseed.enable_aovs:
+            self.__emit_line('<aovs>')
```
dictoon: Use `__open_element()` and `__close_element()` whenever possible.

properties/scene.py
```diff
@@ -49,11 +49,54 @@ def camera_enumerator(self, context):
...
+                                              ('linear', "Linear", ""),
+                                              ('sRGB', "sRGB", ""),
+                                              ('Rec709', "Rec.709", "")],
+                                         default='sRGB')
+
+    output_depth = bpy.props.EnumProperty(name="Output Bit Depth",
+                                          description="The bit depth of the output file.  Leave at default for no conversion",
+                                          items=[
+                                              ('default', "Default", ""),
+                                              ('uint8', "Uint8", "Unsigned 8 bit integer"),
```
dictoon: Add hyphens: 8-bit, 16-bit, etc.

properties/scene.py
```diff
@@ -447,12 +490,47 @@ def update_stamp(self, context):
                                             step=3,
                                             precision=3)
 
+    # AOV export
+
+    enable_aovs = bpy.props.BoolProperty(name="enable_aovs",
+                                         description="Enabled additional AOV's during rendering",
```
dictoon: *AOVs*

render.py
```diff
@@ -434,14 +474,13 @@ def __process_plane_header(self, process):
         aov_index = aov_header[0]
         aov_name_len = aov_header[1]
         aov_nc = aov_header[2]
-
-        aov_name = os.read(process.stdout.fileno(), aov_name_len)
+        self.aovs[aov_index] = "{0}".format(os.read(process.stdout.fileno(), aov_name_len).decode("utf-8"))
```
dictoon: Is the `"{0}".format()` part actually needed?

render.py
```diff
@@ -461,8 +500,8 @@ def __process_plane_data_chunk(self, process, min_x, min_y, max_x, max_y):
         tile_c = tile_header[5]
 
         # We process only the beauty AOV.
-        if tile_aov_index != 0:
-            return True
+        # if tile_aov_index != 0:
```
dictoon: Why is this code (and the code below) commented out?

operators/__init__.py
jdent02: The filename has to go at the end so a simple .append wont work.  Ditto to using '0' as that would replace the 'maketx' command

operators/__init__.py
jdent02: That's just what I started with :)

operators/__init__.py
jdent02: Not following what you mean here.....

operators/__init__.py
jdent02: D'oh.

operators/__init__.py
jdent02: Double D'oh

projectwriter.py
jdent02: Sorry.  Can't remember why I didn't

render.py
jdent02: Accidentally left it in.

render.py
jdent02: Yes.  For the line

layer = result.layers[0].passes[self.aovs[tile_aov_index]]

The 'passes' part requires an entry with parenthesis ("Combined", "diffuse", etx.)

dictoon: Not sure to follow, this isn't about parentheses but about using `format()` with "{0}" as a format string.

jdent02: I think I see your point but if I remember you wanted to avoid string concatenation.

projectwriter.py
jdent02: On the parentheses, they need to be there in order to use a tuple with the .endswith() argument.  Otherwise you get one of those 'needed 1, given 3' errors.

docs/osl/osl.rst
```diff
@@ -1,2 +1,35 @@
 Open Shading Language
-=====================
\ No newline at end of file
+=====================
+
+What is it?
+-----------
+	Open Shading Language (OSL) is a small, customized computer language intended specifically for writing shaders in a physically plausible rendering engine.  While it was originally designed at Sony Pictures Imageworks for their Arnold renderer it has since been integrated into appleseed and several other renderer engines (such as Renderman and Cycles).
```
dictoon: *RenderMan*, or even better *Pixar's RenderMan*.

docs/osl/osl.rst
```diff
@@ -1,2 +1,35 @@
...
+
+How does it work with appleseed?
+--------------------------------
+	Since the beginning appleseed has had its own native shading system (henceforth the ‘built in system’, or ‘BIS’).  While this system worked well enough, it was limited to defining material surfaces and optionally assigning UV mapped textures to them.  That was it.  It couldn’t do any kind of procedural patterns, coordinate manipulation, fancy BSDF mixing, or any of the other utility functions that were needed for a production renderer.  Instead of expanding the BI system with these features, the decision was made to integrate OSL instead, as it is fully capable of all these features and is developed and hosted by a major visual effects company.  As of the current release (0.8), the BIS system is still available in blenderseed, but it is limited to representing simple materials (although it is currently the only way to use volumetric shading).
```
dictoon: I notice that you're using typographic quotation marks instead of basic ones. I don't think it's a problem but clearly the file is no longer using ASCII encoding. Out of curiosity, what text editor are you using?

dictoon: You define the BIS acronym but then use the BI one.

docs/osl/osl.rst
```diff
@@ -1,2 +1,35 @@
...
+
+How do I use it in blenderseed?
+-------------------------------
+	OSL is available in conjunction with the node editor in blender.  When a material is created by default it will use the BIS system.  To use OSL, click on the ‘Add OSL Nodetree’ button at the top of the material panel.  This will create an OSL nodetree and automatically link it into the material.  This will override all the BIS settings and they will vanish from the panel, and OSL will now be used for this material.  Go into the node editor and select the OSL node tree icon from the bottom menu bar.  Be sure to also select the node tree itself for the material from the same menu.  You can add nodes to the editor using Shift-A.  You can add nodes from the different categories to build up your material, but always make sure the final node is an as_closure2surface node, as that is required to export the material properly.
```
dictoon: Capitalize *Blender*.

dictoon: Shift+A instead of Shift-A

docs/osl/osl.rst
```diff
@@ -1,2 +1,35 @@
...
+What if I decide I don’t want to use OSL?
+-----------------------------------------
+	The BIS system can still be used if desired, just delete the node tree link from the material panel.  If you later decide you want to use OSL after all (because let’s face it, you really do want to use it), you can re-link the node tree to the material.
+
+Where does the OSL shading system get the nodes from?
+-----------------------------------------------------
+	The OSL shaders are found in one of two directories in the appleseed download, “shaders/appleseed” and “shaders/blenderseed”.  When blenderseed starts up it scans those two directories and builds the nodes dynamically based on the parameters contained inside the .oso files.  When the scene is rendered the output file will point to those shaders and describe the connections between them.  The shaders themselves are not copied or moved.
+
+Textures with OSL
+---------------------------
```
dictoon: Underlining too long? Probably has no effect but still.

docs/osl/osl.rst
```diff
@@ -1,2 +1,35 @@
...
+	OSL can directly load most image formats (JPEG, PNG, TIFF).  However, this is not the best practice for using it.  Full sized textures take up a lot of memory in the render, and in many cases objects will not be sufficiently large enough in the output frame to justify the memory space used for high resolution, detailed images.  The solution is to create a .tx file using maketx, a utility that ships with appleseed.  Maketx will take an input image and produce a tiled, mipmapped version that is far more efficient when used with OSL.  This is due to a few different features: for one the tiling allows appleseed to only load the sections of the image that are viewable.  Second the mipmapping allows appleseed to load in the appropriate resolution of the texture depending on how large the object is in the output frame. These two features not only reduce the memory requirement for textures, but it also speeds up the render.
```
dictoon: *Viewable* -> *visible*?

dictoon: "it also speeds up the render" -> "they also speed up the render"

operators/__init__.py
```diff
@@ -150,18 +192,20 @@ def execute(self, context):
 
 
 def register():
-    bpy.utils.register_class(AppleseedAddMatLayer)
-    bpy.utils.register_class(AppleseedRemoveMatLayer)
-    bpy.utils.register_class(AppleseedNewNodeTree)
+    bpy.utils.register_class(AppleseedConvertTextures)
```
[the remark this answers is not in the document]
jdent02: A lot of modules have that backwards.  I'll go through and fix them up

docs/osl/osl.rst
jdent02: Sublime 3

ui/world.py
```diff
@@ -115,7 +115,7 @@ def draw(self, context):
         world = scene.appleseed_sss_sets
 
         row = layout.row()
-        row.template_list("SSS_sets_slots", "", world,
+        row.template_list("SSS_sets_slos SSSSetsProps", "", world,
```
dictoon: Typo: slots, not slos

Issue #226, opened 2018-04-16
Revamp OSL parser 


jdent02:
Redo how parameter adjustment visibility is calculated.

Issue #224, opened 2018-04-16
Rendering Crash


Zaldor:
Blenderseed : 0.8.0 preview 
Appleseed : 1.9.0-beta preview 7

the rendering process stops as soon as the buckets appear 

This error appears in console:
```
Traceback (most recent call last):
  File "C:\Users\Zaldor\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\blenderseed\render.py", line 112, in render
    self.__render_scene(scene)
  File "C:\Users\Zaldor\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\blenderseed\render.py", line 137, in __render_scene
    self.__render_project_file(scene, project_filepath, project_dir)
  File "C:\Users\Zaldor\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\blenderseed\render.py", line 309, in __render_project_file
    if not self.__process_plane_data_chunk(process, min_x, min_y, max_x, max_y):
  File "C:\Users\Zaldor\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\blenderseed\render.py", line 567, in __process_plane_data_chunk
    layer = result.layers[0].passes["Depth"]
KeyError: 'bpy_prop_collection[key]: key "Depth" not found'

location: <unknown location>:-1
```

PR #230, opened 2018-04-17
Bug Fixes

### LINKED ISSUE #229: Refresh texture on texture converter causes Error
Clicking on refresh textures causes an error in blenderseed 0.8.0 texture converter
2018-04-16 23_52_16-blender_ c__users_zaldor_desktop_testscene_uv1 blend
2018-04-16 23_52_45-blender_ c__users_zaldor_desktop_testscene_uv1 blend

jdent02:
#229 
#226 
#224

### DIFF
--- a/util.py
+++ b/util.py
@@ -127,15 +127,17 @@ def create_osl_dict(file, content=None):
             if line.startswith("Default value"):
                 currentElement['default'] = line.split(": ")[-1].replace("\"", "")
                 if "type" in currentElement:
-                    if currentElement["type"] in ["color", "vector", "output vector"]:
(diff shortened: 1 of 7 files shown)


util.py
```diff
@@ -149,13 +151,21 @@ def create_osl_dict(file, content=None):
                     currentElement['type'] = 'intenum'
                     currentElement['options'] = line.split(" = ")[-1].replace("\"", "").split("|")
                     currentElement['connectable'] = False
-                if "min = " in line:
+                if "softmin = " in line:
```
dictoon: That looks very brittle, a single missing or additional space and this will fail. Let's keep it that way for the release since you seem to plan a refactoring/rewrite of the OSL parser anyway.

PR #238, opened 2018-04-20
Fixes

jdent02:


projectwriter.py
```diff
@@ -834,9 +835,11 @@ def __get_osl_node_params(self, node, scene):
         return parameters
 
     def __get_socket_links(self, node):
+        connections = []
         for output in node.outputs:
             if output.is_linked:
-                connections = [node.name, output.socket_osl_id, output.links[0].to_node.name, output.links[0].to_socket.socket_osl_id]
+                for x in range(0, len(output.links)):
```
dictoon: Why not

    for link in output.links:
        ...

?

jdent02: Good question. Using that terminology, ‘link’ is actually a full NodeItem reference, which can’t be used as a list index in the following argument.  Took me some thinking to get around that.

dictoon: > which can’t be used as a list index in the following argument

But then you don't use it as a list index, you use it directly.

jdent02: I can alter it if needed.

dictoon: Sure, just smuggle that into a future PR :)

PR #245, opened 2018-04-21
Manual cover page and titles update

jdent02:
Check out what it will actually look like:

http://blenderseed.readthedocs.io/en/latest/

This is my copy of the docs that is up to date with my master branch

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,6 +3,9 @@ Welcome!
 
 .. image:: /_static/appleseedDocCover_Blender.png
 
+Table of Contents:
+==================
+
 .. toctree::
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -3,6 +3,9 @@ Welcome!
 
 .. image:: /_static/appleseedDocCover_Blender.png
 
+Table of Contents:
```
dictoon: Two options:
- If it's a title, it shouldn't end with `:`
- Or it's not a title and it _can_ end `:`

jdent02: Option A

PR #249, opened 2018-04-24
Improve classes registration/deregistration

dictoon:


dictoon:
I'm not sure it would be a net win because of the added complexity. Sometimes straightforward is better than clever, even if it means a tiny bit of code duplication or manual effort.

PR #254, opened 2018-04-29
Multiple Items

jdent02:


properties/scene.py
```diff
@@ -1,4 +1,3 @@
-
```
dictoon: Revert.

render.py
```diff
@@ -254,13 +254,16 @@ def __render_project_file(self, scene, project_filepath, project_dir=None):
 
         # Launch appleseed.cli.
         threads = 'auto' if scene.appleseed.threads_auto else str(scene.appleseed.threads)
-        cmd = (appleseed_bin_path,
+        cmd = [appleseed_bin_path,
                project_filepath,
                '--to-stdout',
                '--threads', threads,
-               '--message-verbosity', 'warning',
+               '--display-options',
```
dictoon: Remove `--display-options`, it's only useful for debugging.

render.py
```diff
@@ -254,13 +254,16 @@ def __render_project_file(self, scene, project_filepath, project_dir=None):
...
+               '--message-verbosity', 'info',
                '--resolution', str(width), str(height),
-               '--window', str(min_x), str(min_y), str(max_x), str(max_y))
+               '--window', str(min_x), str(min_y), str(max_x), str(max_y)]
+        if scene.appleseed.denoise_mode == 'write_outputs':
+            cmd.extend(('--output', "{0}".format(os.path.join(scene.appleseed.denoise_output_file_name, 'denoise_output.exr'))))
```
dictoon: For consistency, use brackets (i.e. lists) instead of parentheses (i.e. tuples):

    cmd.extend(['--output', "{0}".format(os.path.join(scene.appleseed.denoise_output_file_name, 'denoise_output.exr'))])

dictoon: "{0}".format(s)

is the same as

    s

so I'd suggest to simplify the expression.

docs/osl/osl_metadata.rst
```diff
@@ -7,35 +7,35 @@ When blenderseed scans an OSL shader it looks for several specific metadata tags
...
 - Parameter Metadata:  This is metadata that is used on each parameter.
 	- string label
-		- This defines the name of the parameter that will appear in the UI.
+		This defines the name of the parameter that will appear in the UI.
 	- int connectable
-		- This tells the UI whether it should hide the socket connection for this parameter.  You would use this for any parameter that you do not want to be controllable through a texture. '0' hides the socket.
+		This tells the UI whether it should hide the socket connection for this parameter.  You would use this for any parameter that you do not want to be controllable through a texture. '0' hides the socket.
 	- string help
-		- This is a tool tip that will appear whenever a parameter is hovered over.
+		This is a tool tip that will appear whenever a parameter is hovered over.
```
dictoon: *tooltip* in one word in a CS context.

docs/panels/material/light_material.rst
```diff
@@ -6,21 +6,20 @@ Light Material
...
 - Radiance:
-	- This is the color of the light light_emission
+	This is the color of the light light_emission
 - Profile:
-	- This selects one of two different emission profiles:
-		- Diffuse EDF:
-		- Cone EDF:
-			- This profile allows you to change the angle of distribution, allowing you to create a pseudo spotlight light distribtion from a mesh lamp
+	This selects one of two different emission profiles:
+		Diffuse EDF:
```
dictoon: Description missing?

render.py
jdent02: They are there to make sure parenthesis are around the filename (in case of spaces).

properties/scene.py
```diff
@@ -95,7 +94,7 @@ def update_stamp(self, context):
                                             name="appleseed Texture",
                                             description="")
 
-    textures_index = bpy.props.IntProperty(name="layer_index",
+    textures_index = bpy.props.IntProperty(name="textures_index",
```
dictoon: What is this change about?

jdent02: Making the names consistent

PR #269, opened 2018-05-10
Texture converter enhancement

jdent02:
Pressing the refresh button now removes any textures (and optionally the .tx file) that have been removed from the scene

### DIFF
--- a/operators/__init__.py
+++ b/operators/__init__.py
@@ -28,7 +28,7 @@
 
 import bpy
 import subprocess
-from .. projectwriter import image_extensions
+from ..projectwriter import image_extensions
 from .. import util
 import os
 
@@ -93,7 +93,7 @@ class AppleseedRefreshTexture(bpy.types.Operator):
             texture = collection[texture_index]
             if texture.name not in scene_textures:
                 collection.remove(texture_index)
-                if scene.appleseed.remove_unused_textures:
+                if scene.appleseed.del_unused_tex:
                     converted_texture = os.path.realpath(texture.name.split(".")[0] + '.tx')
                     try:
                         os.remove(converted_texture)
--- a/operators/__init__.py
+++ b/operators/__init__.py
@@ -28,8 +28,9 @@
 
 import bpy
 import subprocess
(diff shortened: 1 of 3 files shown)


operators/__init__.py
```diff
@@ -28,8 +28,9 @@
 
 import bpy
 import subprocess
-from ..projectwriter import image_extensions
+from .. projectwriter import image_extensions
```
dictoon: I think the space here is incorrect.

operators/__init__.py
```diff
@@ -73,17 +74,33 @@ def invoke(self, context, event):
...
-                                num = collection.__len__()
+                                num = len(collection)
                                 collection[num - 1].name = string
 
+        texture_index = len(collection) - 1
+        while texture_index > -1:
+            texture = collection[texture_index]
+            if texture.name not in scene_textures:
+                collection.remove(texture_index)
+                if scene.appleseed.remove_unused_textures:
```
dictoon: Move this test at the beginning of this code block: there's no point in doing anything if `remove_unused_textures` is false.

operators/__init__.py
```diff
@@ -73,17 +74,33 @@ def invoke(self, context, event):
...
+                    converted_texture = os.path.realpath(texture.name.split(".")[0] + '.tx')
+                    try:
+                        os.remove(converted_texture)
+                    except:
+                        print("[appleseed] {0} does not exist".format(converted_texture))
```
dictoon: Maybe make this a warning?

    print("[appleseed] WARNING: {0} does not exist.".format(converted_texture))

operators/__init__.py
jdent02: Maybe remove_unused_textures is misleading.  Whenever the refresh button is clicked it will always remove textures from the list that no longer exist in the scene.  If the 'remove' button is checked then any .tx files for those entries will be deleted as well.  And this op has to be here as it relies on the scene_textures list that is populated by the block above it.

jdent02:
Fixes in.

PR #273, opened 2018-05-12
Add appleseed Python option to OSL parser.

jdent02:
Option is enabled by setting 'APPLESEED_PYTHON_PATH' and "USE_APPLESEED_PYTHON' environment variables.  This also sets the stage for further python development without interfering with existing functionality.

### DIFF
--- a/operators/__init__.py
+++ b/operators/__init__.py
@@ -98,7 +98,7 @@ class AppleseedRefreshTexture(bpy.types.Operator):
                     try:
                         os.remove(converted_texture)
                     except:
-                        self.report[{'ERROR'}: "[appleseed] {0} does not exist".format(converted_texture)]
+                        self.report[{'ERROR'}, "[appleseed] {0} does not exist".format(converted_texture)]
             texture_index -= 1
 
         return {'FINISHED'}
@@ -118,7 +118,7 @@ class AppleseedAddTexture(bpy.types.Operator):
         collection = scene.appleseed.textures
 
         collection.add()
-        num = collection.__len__()
+        num = len(collection)
         collection[num - 1].name = ""
 
         return {'FINISHED'}
@@ -139,7 +139,7 @@ class AppleseedRemoveTexture(bpy.types.Operator):
         index = scene.appleseed.textures_index
 
         collection.remove(index)
-        num = collection.__len__()
+        num = len(collection)
         if index >= num:
             index = num - 1
         if index < 0:
@@ -161,7 +161,7 @@ class AppleseedAddSssSet(bpy.types.Operator):
         collection = world.sss_sets
 
         collection.add()
-        num = collection.__len__()
+        num = len(collection)
         collection[num - 1].name = "SSS Set " + str(num)
 
         return {'FINISHED'}
@@ -180,7 +180,7 @@ class AppleseedRemoveSssSet(bpy.types.Operator):
         index = world.sss_sets_index
 
         collection.remove(index)
-        num = collection.__len__()
+        num = len(collection)
         if index >= num:
             index = num - 1
         if index < 0:
--- a/properties/nodes/oslnode.py
+++ b/properties/nodes/oslnode.py
@@ -92,7 +92,377 @@ class AppleseedOSLNode(Node, AppleseedNode):
         material_node.tree.append(self)
 
 
-def generate_node(node):
+def generate_node_python(node):
+    """
+    Generates a node based on the provided node data
+
+    input format is {name, input properties, output properties, non-connectable properties, filename}
+    properties consist of [name, type, default value, min value, max value].
+    Enum properties have item list in place of min/max
+    """
+
+    def draw_color_float(self, context, node):
+        return 0.5, 0.5, 0.5, 1.0
+
+    def draw_color_color(self, context, node):
+        return 0.8, 0.8, 0.5, 1.0
+
+    def draw_closure_color(self, context, node):
+        return 0.0, 0.8, 0.0, 1.0
+
+    def draw_vector_color(self, context, node):
+        return 0.67, 0.45, 1.0, 1.0
+
+    def draw_point_color(self, context, node):
+        return 0.0, 0.0, 1.0, 1.0
+
+    def draw_uv_color(self, context, node):
+        return 0.6, 0.1, 0.0, 1.0
+
+    def draw_matrix_color(self, context, node):
+        return 1.0, 0.5, 1.0, 1.0
+
+    parameter_types = {}
+    filepaths = []
+    name = node['name']
+    category = node['category']
+    input_sockets = node['inputs']
+    output_sockets = node['outputs']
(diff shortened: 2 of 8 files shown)


util.py
```diff
@@ -79,7 +80,119 @@ def get_osl_search_paths():
...
+    while os.path.basename(appleseed_parent_dir) != 'bin':
+        appleseed_parent_dir = os.path.dirname(appleseed_parent_dir)
+    appleseed_parent_dir = os.path.dirname(appleseed_parent_dir)
+
+    return appleseed_parent_dir
+
+
+def load_appleseed_python_paths():
+    if 'APPLESEED_PYTHON_PATH' in os.environ:
+        print("[appleseed] Python Path set to: {0}".format(os.environ['APPLESEED_PYTHON_PATH']))
```
dictoon: `path`, lower case.

util.py
```diff
@@ -79,7 +80,119 @@ def get_osl_search_paths():
...
+        sys.path.append(os.environ['APPLESEED_PYTHON_PATH'])
+    else:
+        python_path = os.path.join(get_appleseed_parent_dir(), 'lib', 'python2.7')
+        sys.path.append(python_path)
+        print("[appleseed] Python path set to: {0}".format(python_path))
+
+    if platform.system() == 'Windows':
+        bin_dir = get_appleseed_bin_dir()
+        os.environ['PATH'] += os.pathsep + bin_dir
+        print("[appleseed] appleseed.dll path loaded")
```
dictoon: Not sure what this message means: `appleseed.dll` is not a "path", and nothing got loaded here as far as I can see?

util.py
```diff
@@ -79,7 +80,119 @@ def get_osl_search_paths():
...
+
+
+def load_appleseed_python_paths():
+    if 'APPLESEED_PYTHON_PATH' in os.environ:
+        print("[appleseed] Python Path set to: {0}".format(os.environ['APPLESEED_PYTHON_PATH']))
+        sys.path.append(os.environ['APPLESEED_PYTHON_PATH'])
+    else:
+        python_path = os.path.join(get_appleseed_parent_dir(), 'lib', 'python2.7')
+        sys.path.append(python_path)
+        print("[appleseed] Python path set to: {0}".format(python_path))
```
dictoon: Refactor such that the common part (`sys.path.append()` + `print()`) is out of the `if`/`else` block.

util.py
```diff
@@ -79,7 +80,119 @@ def get_osl_search_paths():
...
+    print("[appleseed] Parsing OSL shaders...")
+
+    for shader_dir in shader_directories:
+        if os.path.isdir(shader_dir):
+            print("[appleseed] Searching {0} for OSO files...".format(shader_dir))
+            for file in os.listdir(shader_dir):
+                if file.endswith(".oso"):
+                    print("[appleseed] Reading {0}...".format(file))
+                    filename = os.path.join(shader_dir, file)
+                    d = {}
```
dictoon: Swap this line with the next.

util.py
```diff
@@ -79,7 +80,119 @@ def get_osl_search_paths():
...
+
+    if not get_appleseed_bin_dir():
+        print("[appleseed] WARNING: Path to appleseed's binary directory not set: rendering and OSL features will not be available.")
+        return nodes
+
+    tool_dir, shader_directories = get_osl_search_paths()
+
+    load_appleseed_python_paths()
+
+    import appleseed as asr
```
dictoon: `import` in a function? Not sure if that's a good idea. I don't really know, to be honest.

util.py
dictoon: I think I understand now what you meant here. I would suggest instead to print the value of `os.pathsep + bin_dir`. Something like this would seem more useful:

    appleseed_dll_path = os.pathsep + bin_dir
    print("[appleseed] Path to appleseed.dll determined to be {0}".format(appleseed_dll_path)
    os.environ['PATH'] += appleseed_dll_path

util.py
```diff
@@ -29,6 +29,7 @@
 import multiprocessing
 import os
 import sys
+import platform
```
dictoon: Order alphabetically.

properties/nodes/oslnode.py
```diff
@@ -92,7 +92,372 @@ def traverse_tree(self, material_node):
...
+                    kwargs['soft_min'] = int(soft_minimum)
+                if soft_maximum:
+                    kwargs['soft_max'] = int(soft_maximum)
+
+                setattr(ntype, prop_name, bpy.props.IntProperty(**kwargs))
+
+                parameter_types[prop['name']] = "int"
+
+        elif prop['type'] == 'float':
+
```
dictoon: Remove blank line.

properties/nodes/oslnode.py
```diff
@@ -92,7 +92,372 @@ def traverse_tree(self, material_node):
         material_node.tree.append(self)
 
 
-def generate_node(node):
+def generate_node_python(node):
+    """
+    Generates a node based on the provided node data
+
+    input format is {name, input properties, output properties, non-connectable properties, filename}
```
dictoon: Capitalize: *Input*

operators/__init__.py
```diff
@@ -98,7 +98,7 @@ def invoke(self, context, event):
                     try:
                         os.remove(converted_texture)
                     except:
-                        self.report[{'ERROR'}: "[appleseed] {0} does not exist".format(converted_texture)]
+                        self.report[{'ERROR'}, "[appleseed] {0} does not exist".format(converted_texture)]
```
dictoon: Good catch!

util.py
jdent02: I tried doing the python path load in the top level __init__ file and then imported the appleseed module at the top of the util module. It didn’t work.  I can try messing around with it some more

jdent02: I guess it's against PEP8 and not a good idea.  I'll see if I can get it to work in a conforming way

jdent02: So here's the problem.  The util module is accessed at the very beginning of the __init__ module registration when the preferences module is registered (util. safe_register_class).  The python path set function requires that module to be loaded so it can use the preferences to find the bin paths.  And the appleseed python import doesn't work if those paths aren't set.  So putting the appleseed python import at the top of the utils module fails since the python paths haven't been set yet.  So I pretty much have to put the appleseed python import where it was.

dictoon: I'm not sure to understand that business with the util module, but now I realize that of course, the appleseed Python module can only be found once the search paths for modules have been updated.

dictoon: One more thing: we need to make sure that repeatedly enabling and disabling the plugin still works flawlessly.

est77: An import inside a function is not that uncommon. 
You might want to do it for some reasons, like importing big modules only if needed or importing optional modules like here.

There are alternatives, but local imports are usually cleaner.

Python caches all the imports, calling a function multiple times is not expensive.

jdent02:
Should have covered everything

PR #275, opened 2018-05-23
Fix for export path error

jdent02:


docs/osl/osl_metadata.rst
```diff
@@ -28,6 +28,7 @@ When blenderseed scans an OSL shader it looks for several specific metadata tags
 		
 			- filename (tells blenderseed to add a file selection button to a string property)
 			- null (tells blenderseed to hide the direct control for a parameter)
```
dictoon: Maybe keep `null` at the end of the list? Or at the beginning?

export.py
```diff
@@ -101,7 +101,7 @@ def execute(self, context):
 
         for frame in range(frame_start, frame_end + 1):
             scene.frame_set(frame)
-            file_split = self.filepath.split(".")
+            file_split = os.path.splitext(self.filepath)
```
dictoon: Nice. Did you check if `split(".")` or `split('.')` occurs in other places?

jdent02: I didn't see any.  The only reason it happens here is so the frame number can be inserted into the file name.

PR #279, opened 2018-05-28
Add additional options to texture converter

jdent02:


properties/scene.py
```diff
@@ -62,7 +62,7 @@ class AppleseedTextureConvertProps(bpy.types.PropertyGroup):
                                               ('linear', "Linear", ""),
                                               ('sRGB', "sRGB", ""),
                                               ('Rec709', "Rec.709", "")],
-                                         default='sRGB')
+                                         default='linear')
```
dictoon: Is that change intended? If so, it probably deserves a separate commit.

projectwriter.py
```diff
@@ -801,8 +802,14 @@ def __get_osl_node_params(self, node, scene):
                     parameter = bpy.path.abspath(parameter)
                     if scene.appleseed.sub_textures is True:
                         if parameter.endswith(image_extensions):
-                            base_filename = parameter.split('.')
-                            parameter = "{0}.tx".format(base_filename[0])
+                            image_path_segments = os.path.split(parameter)
```
dictoon: What's `image_path_segments` exactly? You seem to assume that `image_path_segments[0]` is a folder name and `image_path_segments[1]` is a file name?

properties/scene.py
```diff
@@ -88,6 +88,15 @@ def update_stamp(self, context):
 
     # Texture conversion
 
+    tex_output_dir = bpy.props.StringProperty(name="tex_output_dir",
+                                              description="",
+                                              default="",
+                                              subtype='DIR_PATH')
+
+    tex_output_use_cust_dir = bpy.props.BoolProperty(name="tex_output_use_cust_dir",
+                                                         description="",
```
dictoon: Fix indentation.

projectwriter.py
jdent02: It’s the best way I could come up with of saving information about the original texture directory in case a custom one is not used

properties/scene.py
jdent02: Yes. The majority of textures in a physically based material are linear and it was a pain in the ass having to switch so many textures from the old sRGB default

dictoon: Ok. Ideally we would guess the color space from the file extension (HDR file formats such as OpenEXR, HDR, floating point TIFF -> linear; everything else -> sRGB).

Also please make a separate commit for this.

jdent02: The guessing wouldn't work for something like Substance Designer.  It exports everything at the same bit depth.

You should get them to change that :)

jdent02: This part is in a separate commit

projectwriter.py
jdent02: I added comments explaining why those steps are being taken.

dictoon: > I added comments explaining why those steps are being taken.

In the code? I'm not seeing them.

jdent02:
Okay now it should all be up to date.

PR #287, opened 2018-06-15
Add AOVs and revamp motion blur calc

jdent02:


render/tilecallbacks.py
```diff
@@ -110,27 +153,48 @@ def on_tile_end(self, frame, tile_x, tile_y):
...
+        pix = []
+        for y in range(take_y - 1, -1, -1):
+            start_pix = (skip_y + y) * tile_w + skip_x
+            end_pix = start_pix + take_x
+            pix.extend(floats[p * tile_c:p * tile_c + tile_c] for p in range(start_pix, end_pix))
+
+        return pix
+
+    @staticmethod
+    def _format_seconds_to_hhmmss(seconds):
```
est77: The standard python convention is to use one underscore for methods and member variables that you want to use only in subclasses and two underscores for methods and member variables that are private.
This method and other similar ones in this class should start with two underscores.

translators/scene.py
```diff
@@ -191,31 +191,34 @@ def translate_scene(self):
...
+        cam_times = set()
+        obj_times = set()
+        shutter_length = self.bl_scene.appleseed.shutter_close - self.bl_scene.appleseed.shutter_open
+        if self.bl_scene.appleseed.enable_camera_blur:
+            cam_segment = shutter_length / self.bl_scene.appleseed.camera_blur_samples
+            for seg in range(1, self.bl_scene.appleseed.camera_blur_samples):
+                cam_times.update({self.bl_scene.appleseed.shutter_open + (seg * cam_segment)})
+            cam_times.update({self.bl_scene.appleseed.shutter_close})
+            all_times.update(cam_times)
+        if self.bl_scene.appleseed.enable_object_blur:
```
est77: This block is very similar to the camera blur case. We can refactor the common parts into a private method.

translators/scene.py
```diff
@@ -191,31 +191,34 @@ def translate_scene(self):
...
-                    self.bl_scene.frame_set(current_frame, subframe=subframe)
-                    for x in self._object_translators.values():
-                        x.update_transform(subframe)
-                        logger.debug("Sample taken for frame %s, subframe %s" % (current_frame, subframe))
-                    subframe += subframe_step
-                self.bl_scene.frame_set(current_frame)
+        # Determine xform subframes for motion blur
+        all_times = set()
+        cam_times = set()
+        obj_times = set()
```
est77: We can rename obj_times to xform_times to avoid confusion with deform keys when we add them.

translators/scene.py
```diff
@@ -191,31 +191,34 @@ def translate_scene(self):
...
+                obj_times.update({self.bl_scene.appleseed.shutter_open + (seg * obj_segment)})
+            obj_times.update({self.bl_scene.appleseed.shutter_close})
+            all_times.update(obj_times)
+        all_times = sorted(list(all_times))
+        current_frame = self.bl_scene.frame_current
+        for time in all_times:
+            self.bl_scene.frame_set(current_frame, subframe=time)
+            if time in cam_times:
+                for x in self.__camera_translators.values():
+                    x.set_transform_key(time)
```
est77: It can be useful in some translators to also pass cam_times as an argument to set_transform_key.
Same for the other set_transform_key

est77:
There is one thing missing, but we can do it in a following PR if you want.
To support motion blur inside linked groups, we need to implement in the GroupTranslator class:

`
def set_transform_key(self, time):
   ...
`
and inside it, call set_transform_key for each object translator inside the group.
And we need to call the method for each group translator in the scene.

render/tilecallbacks.py
jdent02: https://hackernoon.com/understanding-the-underscore-of-python-309d1a029edc

If I’m reading that right it’s the other way around (single underscore for ‘private’ methods)

est77: When using two underscores for private methods or variables, if you try to use them from outside of the class, you get an error. I think that is safer.

There are cases in which you need subclasses to use a method or variable in a base class, 
but don't want them to be useable from outside of a class (like C++'s protected specifier).
Python does not support that, but it is quite common to use a single underscore for this case. 

When working on the code and refactoring, there are some advantages to have different conventions for both cases.

http://www.diveintopython.net/object_oriented_framework/private_functions.html

jdent02:
Updated

PR #289, opened 2018-06-18
Update blenderseed packaging script

jdent02:


scripts/blenderseed.package.py
```diff
@@ -165,14 +169,15 @@ def __get_required(self, tree, key):
...
+        print("  Path to appleseed source:        " + self.appleseed_source_path)
         print("  Path to appleseed binaries:      " + self.appleseed_bin_path)
         print("  Path to appleseed libraries:     " + self.appleseed_lib_path)
         print("  Path to appleseed shaders:       " + self.appleseed_shaders_path)
         print("  Path to appleseed schemas:       " + self.appleseed_schemas_path)
         print("  Path to appleseed settings:      " + self.appleseed_settings_path)
         print("  Path to appleseed.python:        " + self.appleseed_python_path)
         print("  Path to maketx:                  " + self.maketx_path)
-        print("  Output directory:                " + self.package_temp_dir)
+        print("  Output directory:				  " + self.output_dir)
```
dictoon: Replace tabs by spaces.

est77:
Probably that is my mistake. I think I added the first scenes.

jdent02:
I’ll fix them up tonight

scripts/blenderseed.package.py
dictoon: This still needs to be fixed.

jdent02: I'm not sure where the problem is.  The alignment looks correct in both PyCharm and Sublime (i.e. looks like the red line instead of the green)

dictoon: It's using tabs instead of spaces.

dictoon: You can see the problem here on GitHub.

jdent02: Right, I can see it here but it looks correct in my editors, even after doing a PEP8 pass.

dictoon: It appears that you fixed the issue in the meantime, so nevermind and sorry for the noise!

PR #303, opened 2018-08-04
Doc updates

jdent02:


docs/about.rst
```diff
@@ -3,13 +3,11 @@ About blenderseed
...
-	- Contributing Authors:
-		- François Beaune
-		- Luke Kliber
-		- Jonathan Dent
-		- Joel Daniels
-		- Esteban Tovagliari
-		- Jasper van Nieuwenhuizen
-		- Petra Gospodnetic
\ No newline at end of file
+    - Contributing Authors:
```
est77: We usually use alphabetic order for contributors.

est77: We usually use alphabetic order for contributors.

PR #306, opened 2018-08-10
Mesh translators work

est77:
- Do not instance modified meshes.
- Fix infinite loop in insert_entity_with_unique_name.
- Fix meshes names when exporting projects.
- Misc fixes and cleanups.


translators/mesh.py
```diff
@@ -239,20 +208,20 @@ def flush_entities(self, assembly):
             # We always create assemblies when doing IPR to allow quick scene edits.
             needs_assembly = True
         else:
+            # Only create an assembly if the object is intanced of has xform motion blur.
```
jdent02: Typo

est77: Fixed and rebased. Thanks!

PR #308, opened 2018-08-13
Auto node tree and render stats update

jdent02:


render/tilecallbacks.py
```diff
@@ -155,8 +159,8 @@ def on_tile_end(self, frame, tile_x, tile_y):
         seconds_per_pixel = (time.time() - self.__time_start) / self.__rendered_pixels
         remaining_seconds = (self.__total_pixels - self.__rendered_pixels) * seconds_per_pixel
         self.__rendered_tiles += 1
-        self.__engine.update_stats("appleseed Rendering: Pass %i of %i, Tile %i of %i" % (self.__pass_number, self.__total_passes, self.__rendered_tiles, self.__total_tiles),
-                          "Time Remaining: {0}".format(self.__format_seconds_to_hhmmss(remaining_seconds)))
+        self.__render_stats = ["appleseed Rendering: Pass %i of %i, Tile %i of %i" % (self.__pass_number, self.__total_passes, self.__rendered_tiles, self.__total_tiles), "Time Remaining: {0}".format(self.__format_seconds_to_hhmmss(remaining_seconds))]
+        self.__engine.update_stats(self.__render_stats[0], self.__render_stats[1])
```
est77: We are calling update_stats twice from 2 different places.
Since RendererController.get_status is called more often, I think we can move the responsibility to call update_stats to the RendererController and remove the update_stats call here.

PR #320, opened 2018-09-11
Dupli objects support

est77:
- Add support for duplis
- Motion blur related fixes
- Code cleanups


jdent02:
I'm going to go ahead and merge this as is.  We can fix the interactive update later.

translators/dupli.py
```diff
@@ -0,0 +1,191 @@
+#
```
dictoon: Start files with an empty line.

translators/dupli.py
```diff
@@ -0,0 +1,191 @@
...
+# This source file is part of appleseed.
+# Visit http://appleseedhq.net/ for additional information and resources.
```
dictoon: The URL should start with `https://`. Make sure this is the case for all files.

translators/scene.py
```diff
@@ -508,9 +519,12 @@ def __calc_motion_subframes(self):
         self.bl_scene.frame_set(current_frame)
 
     def __get_subframes(self, shutter_length, samples):
+        assert(samples > 1)
```
dictoon: `assert` in Python is an operator, not a function:

    assert samples > 1

translators/camera.py
```diff
@@ -209,11 +209,19 @@ def __thin_lens_camera_params(self, scene, aspect_ratio, film_width, film_height
 
 class InteractiveCameraTranslator(Translator):
 
+    #
```
dictoon: As much as I like these comments, they might not play nice with autopep8. I would suggest to switch to regular docstrings.

PR #323, opened 2018-09-17
Fix for Windows constructed export paths not working on Linux

jdent02:


dictoon:
Worth adding a small comment along the lines of "appleseed files must use Unix path conventions for portability".

est77:
it is not directly related to the PR, but I think we should rename handlers.py to assethandlers.py
Handlers is too generic and does not tell you what are the classes in this file doing.

PR #328, opened 2018-09-22
Tweak packaging script

dictoon:


dictoon:
Ah ok. Then let's remove it.

PR #332, opened 2018-10-12
Package script linux fixes

est77:


scripts/blenderseed.package.py
```diff
@@ -355,7 +355,7 @@ def plugin_extension(self):
 
     def copy_dependencies(self):
         # Create the lib directory.
-        lib_dir = os.path.join(self.settings.package_temp_dir, "lib")
+        lib_dir = os.path.join(self.settings.root_dir, "appleseed", "lib")
```
dictoon: That doesn't look correct. You're creating an `appleseed/bin/` directory in the root _source_ directory of blenderseed?

dictoon: As discussed on Discord, this is exactly as it was meant to be.

Issue #335, opened 2018-10-24
Panoramic camera causes interactive rendering errors


jdent02:
Need to export as perspective camera with an error message (possibly displayed in viewport)

dictoon:
> with an error message (possibly displayed in viewport)

I didn't see an error message in your PR, did I miss it?

PR #336, opened 2018-10-25
Fix for pano camera in interactive mode (#335)

jdent02:


translators/camera.py
```diff
@@ -343,10 +343,10 @@ def __view_camera_params(self, aspect_ratio):
         self.__matrix = self.bl_camera.matrix_world
         cam_mapping = {'PERSP': 'pinhole_camera',
                        'ORTHO': 'orthographic_camera',
-                       'PANO': 'spherical_camera'}
+                       'PANO': 'pinhole_camera'}
```
dictoon: Is this the right thing to do, or just a workaround to avoid crashing or to avoid incorrect results?

PR #359, opened 2018-11-21
Add path tracing overview to docs

jdent02:


docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
+.. title:: Path Tracing Tips
+
+Path Tracing Tips 
+=================
+i.e. how do all these crazy parameters fit together?
+
+|
+
+What is Path Tracing, and why should I care?
+    Path tracing is one of the most dominant (if not the dominant) rendering technique in use today, and the images it produces can be almost indistinguishable from reality.  It does this by using raytracing to accurately trace the path a beam of light takes as it bounces, reflects and refracts its way to the camera lens.  While computationally expensive to do, path tracing has the advantage of naturally capturing the way light moves through a scene.  Effects like global illumination (illumination created by light reflecting off objects) occur naturally with path tracing.  With older methods such as REYES rasterization, these effects had to be laboriously faked with dozens (or hundreds) of manually placed point lights.
```
dictoon: Remove *most* in *one of the most dominant*.
Emphasize *the* (with italic) in *if not the dominant*.
Remove *to do* in *While computationally expensive to do*.
Remove *rasterization* in *REYES rasterization*.

virtualritz: I really like that you are skimming your docs for weasel words. :)

Replace
"With older methods such as REYES rasterization, these effects had to be laboriously faked with dozens (or hundreds) of manually placed point lights."
with
"With older methods these effects were laborious to create."

It's more honest.

My [long] comment as someone who worked in production during the height of REYES use in blockbuster VFX is this:
Most REYES renderers were hybrid ray tracers from around 2000 onwards. They had stuff like occlusion() and indirectdiffuse() shadeops etc. already. These used ray tracing and path tracing.

Then there were point based techniques and shading hacks. Multiple shadow maps for faking area light soft shadows was a PRMan exclusive. It was never laborious to use. You just specified a size of your light source and the number of shadow maps. The rest was automatic.

Multiple (ray-traced) point lights (for e.g. faking ambient occlusion) was stuff of the late 1990s in BMRT which was a ray tracer. I'm not aware of anyone ever using this technique in production either (much too slow at the time).
The famous 'Production Ready Global Illumination' paper from ILM is from 2002 and all the fake stuff referred to in there already uses ray tracing.
Again, none of that stuff was laborious. Now the point (cloud) based stuff that came a few years after – that really made pipelines complex. Good riddance. :)


Mind you, path tracing is also fake. It doesn't even get close to what happens in nature. It just happens to look better than what we did before. :-P
And I'm happy to bet it will look worse (more subtly so, but still) than whatever will be 'state of the art' in a few years from now.

dictoon: Thanks for chiming in Moritz!

What's really stricking to me is the amount of tricks and fakery that path tracing still requires when you need to render complex interiors in production. All kind of knobs need to be adjusted to kill fireflies, and I've even been told by artists that they revert to (again!) manually placing point lights to fake indirect light, etc. 

Depressing :)

dictoon:
@jdent02 May I suggest a separate PR for the commit "Move world settings to Blender world datablocks"?

docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
...
+
+What are you skipping?
+    Quite a bit.  While the method described above is the path an individual ray takes, a single sample per pixel would account for only a miniscule fraction of the possible paths light could take through the scene to that pixel.  This lack of accuracy is visible as noise in the image.  Path tracers get around this limitation by shooting dozens or even hundreds of rays through each pixel.  As each ray follows a slightly different path, the color of the pixel will gradually get closer and closer to ‘converged’, or correct.  As this happens the noise in the image also decreases.  While a path tracer can never perfectly converge a pixel or image (the samples needed is almost infinitely high), eventually it gets close enough that the remaining noise is not objectionable to the human eye.
```
dictoon: Reformulate "While the method described above is the path an individual ray takes"

Reformulate "will gradually get closer and closer to ‘converged’, or correct" -> maybe "will converge, that is, gradually get closer and closer to the correct value" (but in proper English).

docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
...
+
+What else?
+    Plain path tracing can produce beautiful images, eventually.  One challenge that is frequently encountered is that light sources aren’t found often enough when rays bounce around randomly (especially small sources).  To solve this issue, when a surface is hit by a ray, a secondary path (or paths) will be traced from that point directly to a light source.  This is often referred to as ‘next event estimation’ (at least by appleseed).  One of the major features that distinguish one path tracer from another is how they perform this direct light sampling in scenes with dozens (or hundreds) of lights.
```
dictoon: "‘next event estimation’ (at least by appleseed)" -> by appleseed and in the scientific literature.

docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Okay, but path tracing is soooooooo slow
+    Path tracing by its very nature is (and always will be) slower than legacy methods that don’t try to accurately simulate light (such as the REYES method).  The tradeoff is path traced renders naturally achieve realistic lighting that must be laboriously faked in older renders.
```
dictoon: "The tradeoff is path traced renders" -> is that proper English? Just wondering if I've been overcomplicated my sentences all that time :)

docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
...
+
+What are the settings and how do they connect?  Get to the point……
```
dictoon: An ellipsis is three dots, that ought to be enough... :)

docs/general/path_tracing.rst
virtualritz: That depends on the renderer being used. ;) Maybe they're using PRMan which is just in sad state right now?
3Delight does away with all knobs. Spend all developer time on making sure fireflies never show up ... I'm not sure you saw [Aghiles' talk](https://www.youtube.com/watch?v=-ihVkVqCOGc&feature=youtu.be&t=287) on this topic. Worth watching.

docs/general/path_tracing.rst
```diff
@@ -7,52 +7,52 @@ i.e. how do all these crazy parameters fit together?
...
-        This is the control for if light sources are directly sampled at surface hit points.  There’s really no reason why you’d turn this off except if you have no discreet light sources in your scene.  
+        This is the control that determines if light sources are directly sampled at surface hit points.  There’s really no reason why you’d turn this off except if you have no discreet light sources in your scene.
     - The other two controls that affect direct light sampling are the ‘Samples’ and “Low Light Threshold’ controls.  Samples sets how many rays are traced to light sources for each surface hit point, so you may want to raise this if you have a scene with a lot of direct lighting.  Low light threshold is a setting that you can use to speed up render convergence at the expense of lighting quality.  It does this by not directly sampling a light if that light is determined to have less illumination on that point than the threshold setting.  It’s a bit of a tradeoff, less light rays mean faster convergence, but too high of a threshold can cause the lighting of the entire scene to darken.
 
-    - Image Based Lighting: 
+    - Image-Based-Lighting:
```
dictoon: It's really "Image-Based Lighting", the only hyphen is between Image and Based.

dictoon:
Thanks! Really nice docs.

docs/general/path_tracing.rst
```diff
@@ -0,0 +1,81 @@
...
+Path Tracing Tips 
+=================
+i.e. how do all these crazy parameters fit together?
+
+|
+
+What is Path Tracing, and why should I care?
+    Path tracing is one of the dominant (if not *the* dominant) rendering technique in use today, and the images it produces can be almost indistinguishable from reality.  It does this by using raytracing to accurately trace the path a beam of light takes as it bounces, reflects and refracts its way to the camera lens.  While computationally expensive, path tracing has the advantage of naturally capturing the way light moves through a scene.  Effects like global illumination (illumination created by light reflecting off objects) occur naturally with path tracing.  With older methods such as REYES, these effects were laborious to create.
+
+Path Tracing explained in thirty seconds……
```
dictoon: Another place where two full ellipsis are used instead of one (3 dots).

PR #360, opened 2018-11-21
blenderseed C++ export

jdent02:


dictoon:
Approved, thanks! May I suggest to make the commit titles less ambiguous, maybe:

    Add C++ geometry exporter
    Remove Python geometry exporter

?

jdent02:
Done.

PR #365, opened 2018-11-26
Bug fixes and small additions

jdent02:


preferences.py
```diff
@@ -45,7 +45,21 @@ def update_logger(self, context):
...
+        try:
+            import appleseed as asr
+            box = layout.box()
+            box.label(text=asr.get_synthetic_version_string())
+
+            lib_info = asr.get_third_parties_versions()
+            for key in lib_info:
+                box = layout.box()
+                box.label(text="%s: %s" % (key, lib_info [key]))
+        except:
```
dictoon: Bare `except` statements are somewhat dangerous as they may hide programming errors. Whenever possible, you should only catch exceptions that you are expecting.

ui/render.py
```diff
@@ -25,6 +25,7 @@
 # THE SOFTWARE.
 #
 
+import appleseed as asr
```
dictoon: Is this import statement useful?

translators/assethandlers.py
```diff
@@ -89,42 +87,22 @@ def geometry_dir(self):
...
-            dest_dir = self.textures_dir
-        elif asset_type == AssetType.SHADER_ASSET:
-            dest_dir = self.shaders_dir
-        elif asset_type == AssetType.ARCHIVE_ASSET:
-            dest_dir = os.path.join(self.archives_dir, os.path.splitext(file_name)[0])
-
-        dest_file = os.path.join(dest_dir, file_name)
+            if sub_texture:
+                base_filename = os.path.splitext(file_name)[0]
+                file_name = "{0}.tx".format(base_filename)
```
dictoon: `filename`?

properties/scene.py
```diff
@@ -484,6 +485,10 @@ class AppleseedRenderSettings(bpy.types.PropertyGroup):
                                                min=0,
                                                max=100)
 
+    sppm_pt_max_ray_intensity = bpy.props.FloatProperty(name="sppm_pt_max_ray_intensity",
+                                                                  default=1.0,
```
dictoon: Fix indentation.

translators/camera.py
```diff
@@ -34,7 +34,7 @@
 from .assethandlers import AssetType
 from .translator import Translator
 from ..logger import get_logger
-from ..util import calc_film_dimensions, find_auto_focus_point, get_frame_aspect_ratio
+from ..utils.util import calc_film_dimensions, find_auto_focus_point, get_frame_aspect_ratio
```
dictoon: Autofocus is spelled in one word.

translators/lamps.py
```diff
@@ -34,7 +34,7 @@
 from .translator import Translator, ProjectExportMode
 from .assethandlers import AssetType
 from ..logger import get_logger
-from ..path_util import get_osl_search_paths
+from ..utils.path_util import get_osl_search_paths
```
dictoon: Generally speaking, it's a good practice not to import specific symbols from a module, but rather to just import the module and then prefix symbols from that module with the module name, i.e.

    import os
    os.mkdir("bob")

rather than

    from os import mkdir
    mkdir("bob")

That way the provenance of the symbol (which module it belongs to) is clear.

Probably not something to fix now throughout the codebase, but rather a practice to adopt in new code (and maybe migrate existing code when it gets touched/refactored).

utils/path_util.py
```diff
@@ -61,7 +61,7 @@ def get_appleseed_bin_dir_path():
     if "APPLESEED_BIN_DIR" in os.environ:
         appleseed_bin_dir = os.environ['APPLESEED_BIN_DIR']
     else:
-        appleseed_bin_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "appleseed", "bin")
+        appleseed_bin_dir = os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), "appleseed", "bin")
```
dictoon: Double `os.path.dirname()` is weird, intentional? Maybe join with `..` instead if that's what was intended.

dictoon:
Awesome!

Is "Fix world environment update" a fix for an earlier commit of the PR, maybe "Move environment settings to world data block"? If so, please merge them. Otherwise it's fine as is.

utils/path_util.py
jdent02: Yes, intentional as it has to go up two directories to get to the correct starting folder.

preferences.py
jdent02: The preferences file needs to register first at startup as the logger settings are found there.  But then that draw code tries to execute before the environment has been modified with appleseed's location so it fails.  So I need to delay that code running until after that path modification.  I'll try to shuffle things around.

ui/render.py
jdent02: It's not there anymore

jdent02:
Everything should be addressed

PR #372, opened 2018-11-30
Add noise_seed/photon_target

jdent02:


translators/scene.py
```diff
@@ -647,13 +647,17 @@ def __translate_frame(self):
             width = int(self.bl_scene.render.resolution_x * scale)
             height = int(self.bl_scene.render.resolution_y * scale)
 
+        noise_seed = asr_scene_props.noise_seed + self.bl_scene.frame_current \
```
dictoon: I suggest adding parentheses to make operator precedence unambiguous. It's not obvious to me right now how this expression is interpreted.

dictoon: It's interesting that you're still considering the noise seed even when "Vary by Frame" is checked.

properties/scene.py
```diff
@@ -199,6 +199,14 @@ class AppleseedRenderSettings(bpy.types.PropertyGroup):
                                           description="Automatically determine the number of rendering threads",
                                           default=True)
 
+    noise_seed = bpy.props.IntProperty(name="noise_seed",
+                                       default=0)
+
+    per_frame_noise = bpy.props.BoolProperty(name="per_frame_noise",
+                                             description="Vary the noise seed by the frame number.\n"
+                                                         "This gives better results during animations.",
```
dictoon: *in* animations maybe?

translators/camera.py
```diff
@@ -137,7 +137,7 @@ def __set_params(self, scene):
     def __ortho_camera_params(self, scene, aspect_ratio):
         camera = self.bl_camera.data
         cam_params = {'aspect_ratio': aspect_ratio,
-                      'near_z': camera.appleseed.near_z,
+                      'near_z': camera.clip_start,
```
dictoon: Are you sure that `camera.clip_start` is negative? In appleseed, Z+ goes out of the screen (toward the user) while Z- goes into the screen (away from the user). In the local space of the camera, we want the near Z plane to be slightly away from the origin along Z-, hence it must be a negative value.

ui/__init__.py
```diff
@@ -157,6 +115,9 @@ def register():
     #    particles.register()
     lamps.register()
 
+    for panel in get_panels():
+        panel.COMPAT_ENGINES.add('APPLESEED_RENDER')
```
dictoon: Doesn't this need to be undone in `unregister()`?

ui/textures.py
```diff
@@ -0,0 +1,124 @@
+
+#
+# This source file is part of appleseed.
+# Visit https://appleseedhq.net/ for additional information and resources.
+#
+# This software is released under the MIT license.
+#
+# Copyright (c) 2014-2018 Jonathan Dent, The appleseedhq Organization
```
dictoon: New file hence notice should only contain today's date.

Maybe at some point we'll revisit and simplify our rules for copyright notices...

ui/textures.py
```diff
@@ -0,0 +1,124 @@
...
+            if tex:
+                split = layout.split(percentage=0.2)
+                split.label(text="Type:")
+                split.prop(tex, "type", text="")
+
+
+class AppleseedTextureMapping(AppleseedTexturePanelBase):
+    bl_label = "Mapping"
+    bl_context = "texture"
+
```
dictoon: Remove one blank line?

ui/textures.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def draw(self, context):
+        layout = self.layout
+
```
dictoon: Remove blank line?

ui/textures.py
```diff
@@ -0,0 +1,124 @@
...
+        node = context.texture_node
+
```
dictoon: Remove blank line?

ui/textures.py
```diff
@@ -0,0 +1,124 @@
...
+def register():
+    util.safe_register_class(AppleseedTexturePanelBase)
+    util.safe_register_class(AppleseedTextureContext)
+    util.safe_register_class(AppleseedTextureMapping)
+
+
+def unregister():
+    util.safe_unregister_class(AppleseedTextureMapping)
+    util.safe_unregister_class(AppleseedTextureContext)
+    util.safe_unregister_class(AppleseedTexturePanelBase)
```
dictoon: Make sure the last line ends with an end-of-line character.

translators/scene.py
jdent02: Is that not the way to do it?

dictoon: I don't know if it's _the_ way, in any case it's better than what I had in mind.

jdent02:
Everything's fixed I think

translators/camera.py
```diff
@@ -137,7 +137,7 @@ def __set_params(self, scene):
     def __ortho_camera_params(self, scene, aspect_ratio):
         camera = self.bl_camera.data
         cam_params = {'aspect_ratio': aspect_ratio,
-                      'near_z': camera.clip_start,
+                      'near_z': camera.appleseed.near_z,
```
dictoon: Ok, that's one way to do it, however I wasn't against using `camera.clip_start` (but negated) if it makes sense...

translators/scene.py
```diff
@@ -625,7 +625,10 @@ def __translate_render_settings(self):
 
         if asr_scene_props.shading_override:
             parameters['shading_engine'] = {'override_shading': {'mode': asr_scene_props.override_mode}}
+            
         conf_final.set_parameters(parameters)
+
+        parameters['lighting_engine'] = 'pt'
```
dictoon: That looks a bit like a hack :)

I don't quite understand how this meshes with

    lighting_engine = asr_scene_props.lighting_engine if self.export_mode != ProjectExportMode.INTERACTIVE_RENDER else 'pt'

a few lines above.

Moreover, all settings (PT *and* SPPM) should always be exported, even when just one engine is selected. That allows for instance to change the render engine in appleseed.studio and still benefit from the parameters set in Blender. Let me know if I'm not clear, it's early morning here :)

On an unrelated note, the following code (lines 604-608) can be set via the dictionary like other parameters:

```
parameters['pt']['max_diffuse_bounces'] = asr_scene_props.max_diffuse_bounces if not asr_scene_props.max_diffuse_bounces_unlimited else -1
parameters['pt']['max_glossy_bounces'] = asr_scene_props.max_glossy_brdf_bounces if not asr_scene_props.max_glossy_brdf_bounces_unlimited else -1	             parameters['pt']['max_glossy_bounces'] = asr_scene_props.max_glossy_brdf_bounces if not asr_scene_props.max_glossy_brdf_bounces_unlimited else -1
parameters['pt']['max_specular_bounces'] = asr_scene_props.max_specular_bounces if not asr_scene_props.max_specular_bounces_unlimited else -1	             parameters['pt']['max_specular_bounces'] = asr_scene_props.max_specular_bounces if not asr_scene_props.max_specular_bounces_unlimited else -1
parameters['pt']['max_volume_bounces'] = asr_scene_props.max_volume_bounces if not asr_scene_props.max_volume_bounces_unlimited else -1	             parameters['pt']['max_volume_bounces'] = asr_scene_props.max_volume_bounces if not asr_scene_props.max_volume_bounces_unlimited else -1
parameters['pt']['max_bounces'] = asr_scene_props.max_bounces if not asr_scene_props.max_bounces_unlimited else -1	             parameters['pt']['max_bounces'] = asr_scene_props.max_bounces if not asr_scene_props.max_bounces_unlimited else -1
```

dictoon: Ok, thanks.

dictoon:
What's the status of this? Do you have more changes planned @jdent02 or should we merge it?

jdent02:
I do have more stuff I’m working on but this can be merged and I’ll do a new PR

dictoon:
Ok, merging now.

Please don't forget about my remark "On an unrelated note, the following code (lines 604-608) can be set via the dictionary like other parameters" :)

dictoon:
Ah, before I merge, can you squash the feedback commit please?

jdent02:
I’ll do that tonight

translators/scene.py
```diff
@@ -598,14 +598,14 @@ def __translate_render_settings(self):
                                 'volume_distance_samples': asr_scene_props.volume_distance_samples,
                                 'dl_light_samples': asr_scene_props.dl_light_samples,
                                 'ibl_env_samples': asr_scene_props.ibl_env_samples,
-                                'dl_low_light_threshold': asr_scene_props.dl_low_light_threshold}
+                                'dl_low_light_threshold': asr_scene_props.dl_low_light_threshold,
+                                'max_diffuse_bounces': asr_scene_props.max_diffuse_bounces if not asr_scene_props.max_diffuse_bounces_unlimited else -1,
+                                'max_glossy_bounces': asr_scene_props.max_glossy_brdf_bounces if not asr_scene_props.max_glossy_brdf_bounces_unlimited else -1,
```
dictoon: Note for later: rename `max_glossy_brdf_bounces`, it's the only one containing the word `brdf`.

Issue #377, opened 2018-12-04
Modify packaging script version labeling


jdent02:
use 'git describe --long' for version, followed by platform, followed by date.

Extra credit: automate date using system time module.

dictoon:
Woops misclick :)

PR #379, opened 2018-12-05
Update packaging script with correct label formatting (#377)

jdent02:


scripts/blenderseed.package.py
```diff
@@ -324,7 +326,7 @@ def clean_stage(self):
 
     def build_final_zip_file(self):
         progress("Building final zip file from staging directory")
-        package_name = "blenderseed-{0}-{1}".format(self.package_version, self.settings.platform)
+        package_name = "blenderseed-{0}-{1}_{2}".format(self.package_version, self.settings.platform, self.build_date)
```
dictoon: Replace underscore by dash, let's not mix them :)

scripts/blenderseed.package.py
```diff
@@ -690,13 +692,15 @@ def main():
...
     parser.add_argument("--nozip", action="store_true", help="copies appleseed binaries to blenderseed folder but does not build a release package")
 
     args = parser.parse_args()
 
     no_release = args.nozip
-    package_version = args.version if args.version else "no-version"
+    package_version = subprocess.Popen("git describe --long", stdout=subprocess.PIPE, shell=True).stdout.read().strip()
+
+    date_time = datetime.datetime.now()
+    build_date = "%i-%i-%i" % (date_time.year, date_time.month, date_time.day)
```
dictoon: That's not correct, February will be printed as 2 instead of 02, messing sorting order up.

You can use `date.today().isoformat()` which will return `2018-12-05`.

scripts/blenderseed.package.py
```diff
@@ -690,13 +692,15 @@ def main():
...
     parser = argparse.ArgumentParser(description="build a blenderseed package from sources")
 
-    parser.add_argument("--version", help="version number of packaged file")
     parser.add_argument("--nozip", action="store_true", help="copies appleseed binaries to blenderseed folder but does not build a release package")
 
     args = parser.parse_args()
 
     no_release = args.nozip
-    package_version = args.version if args.version else "no-version"
+    package_version = subprocess.Popen("git describe --long", stdout=subprocess.PIPE, shell=True).stdout.read().strip()
```
dictoon: Is `args.version` still used?

jdent02: No

scripts/blenderseed.package.py
jdent02: Done

PR #387, opened 2018-12-12
Bug fixes and updates

jdent02:


translators/object.py
```diff
@@ -118,6 +120,13 @@ def flush_entities(self, assembly):
     def update_transform(self, time, matrix):
         self.__ass_inst.transform_sequence().set_transform(time, self._convert_matrix(matrix))
 
+    def _convert_matrix(self, m):
```
dictoon: Private methods must start with two underscores.

translators/object.py
```diff
@@ -161,3 +170,9 @@ def flush_entities(self, assembly):
 
     def update_transform(self, time, matrix):
         self.__ass_inst.transform_sequence().set_transform(time, self._convert_matrix(matrix))
+
+    def _convert_matrix(self, m):
```
dictoon: Same here.

translators/object.py
dictoon: Do I understand correctly that this method not only returns a new matrix, but also modifies the argument that is passed to it? Is this modification visible to the outside? I.e.

	m1 = some matrix
	m2 = self._convert_matrix(m1)
	# is m1 different now?

jdent02: It’s not private.  It’s an override for the same function in the base Translator class

jdent02: It’s an override for the same function in the base Translator class.  Esteban said those have single underscores. Only internal functions which are not inherited need double underscores.  This convention is followed throughout blenderseed.

dictoon: Thanks for the explanation. Today I learn!

jdent02: I think the single underscore is the closest thing Python has to a 'protected' member.

PR #388, opened 2018-12-12
Refactor node UI

jdent02:


jdent02:
This one is ready for review I think

jdent02:
I don't think you're looking at the latest revision.....

dictoon:
Hm, I just checked and yes I'm looking at the latest revision... :) Which comments aren't relevant in your opinion?

jdent02:
These ones:
image

jdent02:
This one:

image

The file actually looks like this:

```python
for y in socket_input_names:
                input = self.inputs.new(y['socket_name'], y['socket_label'])
                if (y['connectable'] is True and y['hide_ui'] is not True) or y['connectable'] is False:
                    input.hide = True
```

jdent02:
As far as the socket color question, I'm hamstrung by the way Blender works.  Each socket class needs a
```python
def draw_color(self, context, node)
```
function.  It can only take those arguments.  I could factorize it probably but it would be much more convoluted.  The way I have takes up more screen space but it's conceptually more straightforward.

jdent02:
I've addressed your comments on the 'x' and 'y' variables not being well named.

dictoon:
Can you please merge related commits together?

jdent02:
I'll work on it on a bit.  I have an addition or two to make.

jdent02:
Oh, and to answer your backwards compatibility question: from all of my tests backwards compatibility is maintained.  At file load a new util function called 'update_project' runs.  It copies any modified socket values into the identical value in the node.  In the future this util function could also make other updates if we change additional things.

properties/nodes/oslnode.py
```diff
@@ -164,7 +164,7 @@ def draw_buttons(self, context, layout):
         socket_number = 0
         param_section = ""
         for x in input_params:
```
dictoon: Another place where `x` is used for iteration.

PR #391, opened 2018-12-18
Bug Fixes

jdent02:


translators/dupli.py
```diff
@@ -159,6 +164,17 @@ def __create_dupli_instances(self, obj, scene, settings):
 
                         self.__object_translators.append(translator)
 
+                        for slot in dupli.object.material_slots:
+                            mat = slot.material
+                            mat_key = ObjectKey(mat)
+
+                            if mat_key not in self.__material_translators:
+                                logger.debug(
```
dictoon: Write statement on a single line?

translators/dupli.py
```diff
@@ -67,6 +68,7 @@ def __init__(self, dupli, export_mode, asset_handler):
         self.__settings = 'VIEWPORT' if self.__export_mode == ProjectExportMode.INTERACTIVE_RENDER else 'RENDER'
 
         self.__object_translators = []
```
dictoon: Whys is this an array instead of a dictionary? It seems to be used as a dictionary...

translators/group.py
```diff
@@ -211,25 +211,28 @@ def _create_translators(self, scene):
...
-                            logger.debug("Saving translator for object %s in instance map", obj_key)
-                            self._datablock_to_translator[mesh_key] = translator
+                            if not is_modified:
+                                logger.debug("Saving translator for object %s in instance map", obj_key)
+                                self._datablock_to_translator[mesh_key] = translator
 
-                        self.__create_material_translators(obj)
+                            self.__create_material_translators(obj)
+                    else:
+                        logger.debug("Error: Mesh %s has no polygons, skipping export", mesh_key)
```
dictoon: Either that's an error and then you shoulnd't use `logger.debug()`, or it's a debug message and you shouldn't prefix the message with "Error", or (my preferred choice) it's a simple info message.

translators/camera.py
```diff
@@ -34,7 +34,7 @@
 from .assethandlers import AssetType
 from .translator import Translator
 from ..logger import get_logger
-from ..utils.util import calc_film_dimensions, find_autofocus_point, get_frame_aspect_ratio
+from ..utils.util import calc_film_dimensions, find_autofocus_point, get_focal_distance, get_frame_aspect_ratio
```
dictoon: It would be cleaner to import the whole module rather than individual functions, and to prefix function names with their module name.

translators/dupli.py
jdent02: That’s the way Esteban set it up originally.

translators/dupli.py
jdent02: It’s that damn auto formatter in VSCode

translators/camera.py
jdent02: Done

translators/dupli.py
dictoon: I checked more closely and the code is correct, sorry for the noise.

One remark though: it seems that this cleanup code is duplicated in `__clear_dupli_instances()`. Can we factorize it? We're bound otherwise to update one place and not the other...

translators/dupli.py
dictoon: You can probably set the maximum line width. I've set it to 150 here.

translators/dupli.py
jdent02: Possibly.  I think Esteban originally did it this way because of motion blur, as the object dupli list has to be freed and created again for every motion step.

jdent02: Seems to work fine without it.

jdent02:
Think I got it all.

dictoon:
What about the test scene reference image? See my question above.

jdent02:
It's been updated, not deleted.  Look at the diff for the file

translators/dupli.py
jdent02: The redundant code in __clear_dupli_instances()

dictoon:
Ow my bad, got again tricked by GitHub's poor UI:

image

Good to merge then, please rebase and merge relevant commits together.

Thanks!

translators/dupli.py
jdent02: I know, I just haven't done it yet :)

PR #402, opened 2019-02-27
Fix for node frames

jdent02:


translators/materials.py
```diff
@@ -176,9 +178,10 @@ def _reset(self, mat):
     def __set_shader_group_parameters(self, scene):
         surface_shader = None
         for shader in self.__shaders:
-            if shader.node_type == 'osl_surface':
-                surface_shader = shader
-                self.__shader_list = surface_shader.traverse_tree()
+            if not isinstance(shader, bpy.types.NodeInternal):
+                if shader.node_type == 'osl_surface':
+                    surface_shader = shader
+                    self.__shader_list = surface_shader.traverse_tree()
```
dictoon: Shouldn't you break the `for` loop when you found what you were looking for?

render/__init__.py
```diff
@@ -235,6 +235,7 @@ def __start_final_render(self, scene, project):
 
         self.__renderer = asr.MasterRenderer(project,
                                              project.configurations()['final'].get_inherited_parameters(),
+                                             [],
```
dictoon: How did it work until now?

translators/materials.py
jdent02: Probably....

render/__init__.py
jdent02: Master hasn't been updated in a while.  The last time was before Esteban's OSL source code merge, which is when that additional 'resource path' was added.

dictoon: Oh ok, got it. Thanks.

translators/materials.py
dictoon: Do you want to fix this before merging the PR?

jdent02:
Yep. Should get to it today

PR #415, opened 2019-04-25
Mesh conversion function update (and other fixes)

jdent02:


translators/scene.py
```diff
@@ -625,10 +625,11 @@ def __translate_render_settings(self):
                                   'max_path_length': 0,
                                   'max_photons_per_estimate': asr_scene_props.sppm_max_per_estimate,
                                   'path_tracing_max_path_length': asr_scene_props.sppm_pt_max_length,
-                                  'path_tracing_max_ray_intensity': asr_scene_props.sppm_pt_max_ray_intensity,
                                   'path_tracing_rr_min_path_length': asr_scene_props.sppm_pt_rr_start,
                                   'photon_tracing_max_path_length': asr_scene_props.sppm_photon_max_length,
                                   'photon_tracing_rr_min_path_length': asr_scene_props.sppm_photon_rr_start}
+            if asr_scene_props.sppm_pt_max_ray_intensity != 0.0:
```
dictoon: In principle 0 is a valid value for MRI. The way this is supposed to work is: if MRI is set, apply it; if it's not set, don't apply it; don't filter by value.

translators/mesh.py
```diff
@@ -336,7 +336,7 @@ def __convert_mesh(self, me):
 
             uv_layer_pointer = active_uv.data[0].as_pointer()
 
-        asr.convert_bl_mesh(self.__mesh_object,
+        asr.convert_tessface_mesh(self.__mesh_object,
                             vertices_length,
```
dictoon: Fix indentation.

translators/scene.py
dictoon: In blenderseed for Blender 2.79, there's an `Unlimited` button next to the MRI input. Here we should check the status of that button, not the value that is set. Does that make sense?

jdent02:
Addressed your comments

dictoon:
Good to go, please rename commit "Updates for feedback" or merge it with others.

Issue #414, opened 2019-04-23
Crash after adding/removing post-processing stages


dictoon:
With blenderseed 1.0.0 (latest official release):

I added a stage, removed one, added another... After a short while I got an error:

![image](https://user-images.githubusercontent.com/321290/56607598-3ca9a380-6609-11e9-815e-2e53d28d0551.png)

I also suspect that at some point before the crash one of my Color Map stage got overwritten or replaced by a Render Stamp one.

PR #416, opened 2019-04-29
Fix for issue #414

jdent02:


operators/__init__.py
```diff
@@ -295,7 +295,8 @@ def invoke(self, context, event):
             index = num - 1
         if index < 0:
             index = 0
-            scene.post_processing_stages_index = index
+
+        scene.post_processing_stages_index = index
```
dictoon: I feel this is an argument against using local copies of variables. If `scene.post_processing_stages_index` had been used directly, the bug probably wouldn't have occurred:

    scene.post_processing_stages_index = clamp(scene.post_processing_stages_index, 0, len(collection) - 1)

with `clamp(x, low, high)` being something like min(low, max(high, x))`.

For `collection`, it's different since it's a reference to the original one (`scene.post_processing_stages`), unlike for `index` which is a value copy.

PR #434, opened 2019-07-29
Small Fixes

Mango-3:
- Fix Albedo AOV
- Remove obsolete pixel filter

### DIFF
--- a/translators/scene.py
+++ b/translators/scene.py
@@ -667,6 +667,8 @@ class SceneTranslator(object):
             logger.debug("Translating AOVs")
             asr_scene_props = self.bl_scene.appleseed
 
+            if asr_scene_props.albedo_aov:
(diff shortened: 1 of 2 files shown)


translators/scene.py
```diff
@@ -705,6 +705,8 @@ def __set_aovs(self, aovs):
                 aovs.insert(asr.AOV('cryptomatte_material_aov', {}))
             if asr_scene_props.cryptomatte_object_aov:
                 aovs.insert(asr.AOV('cryptomatte_object_aov', {}))
+            if asr_scene_props.albedo_aov:
+                aovs.insert(asr.AOV('albedo_aov', {}))
```
jdent02: Just nitpicking but it would be nice to have this at the top of the list.

Mango-3: Done

PR #459, opened 2019-09-24
Updates and Fixes

jdent02:


utils/util.py
```diff
@@ -112,6 +112,9 @@ def update_project(_):
             for cls in node_classes:
                 safe_register_class(cls)
 
+        else:
+            logger.debug(f"appleseed: Shader {script.name} did not compile")
```
dictoon: Is there a way to include the error message(s)?

translators/objects/mesh.py
```diff
@@ -278,14 +278,14 @@ def __get_material_mappings(self):
 
         if len(material_slots) > 1:
             for i, m in enumerate(material_slots):
-                if m.material.use_nodes is True:
+                if m.material is not None and m.material.use_nodes is True:
```
dictoon: Note that `is True` is redundant.

translators/scene.py
```diff
@@ -529,7 +529,8 @@ def __translate_render_settings(self):
                       'passes': asr_scene_props.renderer_passes,
                       'generic_frame_renderer': {'tile_ordering': asr_scene_props.tile_ordering},
                       'progressive_frame_renderer': {'max_average_spp': asr_scene_props.interactive_max_samples,
-                                                     'max_fps': asr_scene_props.interactive_max_fps},
+                                                     'max_fps': asr_scene_props.interactive_max_fps,
+                                                     'time_limit': asr_scene_props.interactive_max_time},
```
dictoon: Rename `time_limit` to `max_time` for consistency with the other properties?

translators/scene.py
```diff
@@ -676,6 +676,27 @@ def __translate_frame(self, context):
             window_y_min = height - int(camera_height * self.bl_scene.render.border_max_y + window_shift_y - view_shift_y)
             window_y_max = height - int(camera_height * self.bl_scene.render.border_min_y + window_shift_y - view_shift_y)
 
+            #Check for coordinates outside the render window
```
dictoon: Add space after `#`, terminate sentence with a period.

translators/scene.py
```diff
@@ -676,6 +676,27 @@ def __translate_frame(self, context):
...
+            if window_x_min < 0:
+                window_x_min = 0
+            elif window_x_min > width:
```
dictoon: That doesn't look right, for a WxH image, the max cropw window is [0,W-1] x [0,H-1].

I would also suggest introducing a function (if it doesn't already exist) that clamp a value to min and max bounds, i.e. `window_x_max = clamp(window_x_max, 0, width - 1)`.

utils/util.py
jdent02: I haven’t found one. There is an error handler for OIIO tex conversion that’s supposed to print to stdout or stderr but it doesn’t seem to do anything.

translators/scene.py
jdent02: With the window position calculations, if a user shifts the through camera view (using Shift-Middle Mouse) one or more of the border sides could end up outside the view window, leading to a window dimension that is larger than the width/height or a negative number.

translators/objects/mesh.py
jdent02: Fixed

translators/scene.py
jdent02: Can't, that's the name of the appleseed parameter.

translators/scene.py
jdent02: Although a clamp would be an alternative.

translators/scene.py
dictoon: Doh!

jdent02:
Should be fixed.

PR #469, opened 2019-11-17
Add shadow terminator correction

jdent02:


ui/objects.py
```diff
@@ -68,6 +68,7 @@ def draw(self, context):
         layout.prop(asr_obj, "double_sided", text="Double Sided Shading")
         layout.prop(asr_obj, "photon_target", text="SPPM Photon Target")
         layout.prop(asr_obj, "medium_priority", text="Nested Glass Priority")
+        layout.prop(asr_obj, "shadow_terminator_correction", text="Shadow Terminator Correction")
```
dictoon: FWIW it's called Shadow Terminator Fix in appleseed.studio and appleseed-maya.

jdent02:
> FWIW it's called Shadow Terminator Fix in appleseed.studio and appleseed-maya. 

Changed

PR #471, opened 2019-11-21
Small fixes

jdent02:


translators/scene.py
```diff
@@ -437,10 +437,11 @@ def write_project(self, filename):
         Write the appleseed project out to disk.
         :param filename:
         """
+        filename = bpy.path.ensure_ext(filename, '.appleseed')
 
         asr.ProjectFileWriter().write(
             self.as_project,
-            bpy.path.abspath(filename),
```
dictoon: How did that work before exactly?

jdent02: Whoops, I should have left that line in there.  

bpy.path.abspath converts a Blender relative path (beginning with "//") into a full absolute path.

PR #479, opened 2019-12-02
Updates and bug fixes

jdent02:


operators/osl_ops.py
```diff
@@ -38,6 +38,7 @@ class ASMAT_OT_compile_script(bpy.types.Operator):
     bl_label = "Compile OSL Script Node Parameters"
 
     def execute(self, context):
+
```
dictoon: Revert.

operators/osl_ops.py
```diff
@@ -49,73 +50,79 @@ def execute(self, context):
...
+                self.report({'ERROR'}, "appleseed - OSL script did not compile!")
+
+            return {'FINISHED'}
 
         else:
-            self.report({'ERROR'}, "OSL script did not compile!")
+            self.report({'ERROR'}, "appleseed - No OSL script selected!")
 
-        return {'FINISHED'}
+            return {'FINISHED'}
```
dictoon: Factorize this line since it's common to both code paths?

jdent02: Fixed.

jdent02:
This is okay to merge now.

translators/nodetree.py
```diff
@@ -113,11 +113,11 @@ def __create_shadergroup(self, bl_scene):
                     if parameter_value == "int checkbox":
                         parameter_value = "int"
                         parameter = int(parameter)
-                    if parameter_value in ('color', 'vector', 'normal'):
+                    if parameter_value in ('color', 'vector', 'normal', 'float[2]'):
```
dictoon: What would happen with `float[3]` if these were to appear at some point? Could we add a `else` case with a clear error message of what's going on?

Also, shouldn't this be a `elif` instead of just a `if`, for clarity's sake?

translators/nodetree.py
```diff
@@ -113,11 +113,11 @@ def __create_shadergroup(self, bl_scene):
...
                         parameter = " ".join(map(str, parameter))
```
dictoon: What is `parameter`?

It seems that `parameter_value` is really a parameter _type_, in which case it should be called `parameter_type`; and that `parameter` is really a parameter _value_, in which case it should be called `parameter_value`.

utils/osl_utils.py
```diff
@@ -149,52 +154,52 @@ def traverse_tree(self):
         if param['connectable'] is True:
             connectable = True
 
-        if param['hide_ui'] is True:
-            hide_ui = True
+            if param['hide_ui'] is True:
```
dictoon: So all this code is now under `if param['connectable'] is True:`? That's intentional?

I wonder if the code wouldn't be clearer if you changed the first test to:

    if not param['connectable']:
        continue

(and then you can also get rid of the `connectable` variable since it's always true).

properties/nodes.py
```diff
@@ -105,6 +105,8 @@ def draw_buttons(self, context, layout):
                         socket_number += 1
                     else:
                         layout.prop(self, x['name'], text=label_text)
+            else:
+                socket_number += 1
```
dictoon: What's the logic? When do you need to increment or not `socket_number`?

jdent02:
> What would happen with `float[3]` if these were to appear at some point? Could we add a `else` case with a clear error message of what's going on?

Also, shouldn't this be a `elif` instead of just a `if`, for clarity's sake? 

Sorry, deleted your comment accidentally.  This wouldn't really work as there are a number of other parameter types that pass through this function unaffected.  It's only here to change the formatting on some of the parameters so OSL recognizes them properly.

For example, float, integers and strings pass through here without triggering anything.  If I add an 'else' error statement that triggers on all those.

translators/nodetree.py
jdent02: Fixed.

utils/osl_utils.py
jdent02: Yes, it makes sense.  Done.

properties/nodes.py
jdent02: It's incremented whenever a parameter has an associated socket class.  What was happening before is the socket number was not being incremented for pointer types (i.e shader inputs on the asBlendShader) so the split.enabled command for some properties were getting triggered by the wrong socket being connected.

jdent02:
Feedback addressed.

properties/nodes.py
```diff
@@ -138,7 +138,7 @@ def init(self, context):
         if self.socket_input_names:
             for socket in self.socket_input_names:
                 input = self.inputs.new(socket['socket_name'], socket['socket_label'])
-                if (socket['connectable'] is True and socket['hide_ui'] is False) or socket['connectable'] is False:
+                if (socket['hide_ui'] is False):
```
dictoon: So the logic here is different than before, right?

Parentheses are redundant.

translators/nodetree.py
```diff
@@ -102,22 +102,23 @@ def __create_shadergroup(self, bl_scene):
...
-                    parameter = getattr(node, key)
-                    parameter_value = parameter_types[key]
+                    parameter_value = getattr(node, key)
+                    parameter_type = parameter_types[key]
 
                     if key in node.filepaths:
                         sub_texture = bl_scene.appleseed.sub_textures
-                        parameter = self.asset_handler.process_path(parameter.filepath,
+                        parameter_value = self.asset_handler.process_path(parameter_value.filepath,
                                                                     AssetType.TEXTURE_ASSET, sub_texture)
```
dictoon: Fix indentation.

utils/osl_utils.py
```diff
@@ -144,62 +144,61 @@ def traverse_tree(self):
...
-            elif socket_type in ("point", "point[4]"):
-                stype.draw_color = draw_point_color
+        elif socket_type in ("point", "point[4]"):
+            stype.draw_color = draw_point_color
 
-            node_classes.append(stype)
+        node_classes.append(stype)
 
-            socket_input_names.append({'socket_name': socket_name, 'socket_label': socket_label, 'connectable': connectable, 'hide_ui': hide_ui})
+        socket_input_names.append({'socket_name': socket_name, 'socket_label': socket_label, 'hide_ui': hide_ui})
```
dictoon: You're no longer setting `connectable` to `True`, intended?

properties/nodes.py
```diff
@@ -227,10 +230,14 @@ def traverse_tree(self, material_node):
...
                 linked_node = socket.links[0].from_node
-                linked_node.traverse_tree(material_node)
+                if hasattr(linked_node, "traverse_tree"):
+                    linked_node.traverse_tree(material_node)
+                else:
+                    logger.error(f"Node {linked_node.name} is not an appleseed node, stopping traversal")
         material_node.tree.append(self)
 
 
+
```
dictoon: Revert? Pep8 only requires two blank lines.

Talking of which, we should run pep8 again on all files.

properties/nodes.py
```diff
@@ -71,7 +74,7 @@ def traverse_tree(self, material_node):
                 if hasattr(linked_node, "traverse_tree"):
                     linked_node.traverse_tree(material_node)
                 else:
-                    print(f"Node {linked_node.name} is not an appleseed node, stopping traversal")
+                    logger.error(f"Node {linked_node.name} is not an appleseed node, stopping traversal")
```
dictoon: It seems that you are mixing a number of things in this commit, could you split this commit into more focused ones please?

dictoon:
Thanks for all the answers, the PR looks good, just a couple quibbles.

properties/nodes.py
jdent02: Yes, the 'connectable' prop was redundant.

utils/osl_utils.py
jdent02: Yes.  Redundant prop now.

PR #503, opened 2020-04-03
Assorted fixes and tweaks

jdent02:


utils/osl_utils.py
```diff
@@ -318,7 +318,7 @@ def traverse_tree(self):
 
                 ntype.__annotations__[prop_name] = bpy.props.IntProperty(**kwargs)
 
-                parameter_types[param['name']] = "int"
+                parameter_types[param['name']] = param['type']
```
dictoon, line 321: Since this line is now the same in all cases, can it be factorized? (Maybe not, I haven't checked, just asking.)

jdent02: Probably not.  If you look up to line 306 I have to explicitly state the assignment there.

The main reason I did this was to streamline vector, normal and point data since they're all represented by the same Blender property (and they don't require a color widget).

jdent02:
> What is the impact of the changes in commit "Small tweak to RenderEngine"? Looks like a nice simplification, but does it fix anything?

Not really.  Just removes unused variables

PR #515, opened 2020-05-31
Initial support for Cycles nodes

jdent02:


properties/nodes.py
```diff
@@ -74,7 +74,8 @@ def traverse_tree(self, material_node, engine):
                     linked_node.traverse_tree(material_node, engine)
                 else:
                     logger.error(f"Node {linked_node.name} is not an appleseed node, stopping traversal")
-                    engine.report({'ERROR'}, f"Node {linked_node.name} is not an appleseed node, stopping traversal")
+                    if engine is not None:
```
dictoon: What's `engine`?

properties/nodes.py
```diff
@@ -277,6 +263,10 @@ def node_categories(osl_nodes):
     osl_utilities = []
     osl_other = []
 
+    cycles_nodes = list()
```
dictoon: This could be written as a list comprehension (if you're not familiar with that: basically a way to transform a list into another list in a single declarative line; check the web for the details, it's worth knowing).

utils/path_util.py
```diff
@@ -86,6 +86,13 @@ def get_osl_search_paths():
 
     return shader_directories_sorted
 
+def get_cycles_shader_paths():
+    bl_exec = os.path.dirname(bpy.app.binary_path)
+
+    cycles_shader_path = os.path.join(bl_exec, "2.82", "scripts", "addons", "cycles", "shaders")
```
dictoon: Blender version is hardcoded here, not sure we want to do that?

utils/cycles_shader_parsers.py
```diff
@@ -0,0 +1,57 @@
...
+# THE SOFTWARE.
+#
+
+import bpy
+
+from .util import mapping_to_array
+
+
+def parse_cycles_shader(shader):
+
```
dictoon: Remove blank line.

utils/cycles_shader_parsers.py
```diff
@@ -39,19 +41,44 @@ def parse_ShaderNodeRGBCurve(shader):
...
+    map_g = mapping.curves[1]
+    map_b = mapping.curves[2]
+    map_i = mapping.curves[3]
+
+    for i in range(curve_resolution):
+        start_index = i * 3
+        t = i / (curve_resolution - 1)
+        rgb_floats[start_index] = mapping.evaluate(
+            map_r, mapping.evaluate(map_i, t))
+        rgb_floats[start_index +
```
dictoon: Is that how autopep8 formats this code?

utils/cycles_shader_parsers.py
```diff
@@ -26,8 +26,10 @@
 #
 
 import bpy
+import numpy as np
```
dictoon: Is it the first time NumPy is used in blenderseed?

properties/nodes.py
```diff
@@ -247,9 +247,10 @@ def node_categories(osl_nodes):
     osl_utilities = []
     osl_other = []
 
-    cycles_nodes = list()
-    for node in cycles_shader_parsers.cycles_nodes.keys():
-        cycles_nodes.append(nodeitems_utils.NodeItem(node))
+    from ..translators.cycles_shaders import cycles_nodes
```
dictoon, line 250: I don't know if we really want imports in the middle of the code :) I'm not qualified to judge, just asking.

translators/cycles_shaders.py
```diff
@@ -29,7 +29,15 @@
 import numpy as np
 
 cycles_nodes = {"ShaderNodeRGBCurve": "node_rgb_curves.oso",
-                "ShaderNodeValToRGB": "node_rgb_ramp.osl"}
+                "ShaderNodeValToRGB": "node_rgb_ramp.osl",
```
dictoon, line 32: `.osl` here but `.oso` below and above.

translators/cycles_shaders.py
```diff
@@ -30,10 +30,13 @@
 
 cycles_nodes = {"ShaderNodeRGBCurve": "node_rgb_curves.oso",
                 "ShaderNodeValToRGB": "node_rgb_ramp.osl",
-                "ShaderNodeSeparateRGB": "node_separate_rgb.oso"}
+                "ShaderNodeSeparateRGB": "node_separate_rgb.oso",
+                "ShaderNodeCombineRGB": "node_combine_rgb.osl"}
```
dictoon, line 34: And `.osl` here too, not sure why both extensions are used.

properties/nodes.py
jdent02: I have to do it that way to get the module to load. If I put the import at the top of the file it fails to initialize. One of these days I should really refactor the modules so that doesn’t happen

translators/cycles_shaders.py
jdent02: ID10T error from the coder.

properties/nodes.py
jdent02: Traditional representation of bpy.types.RenderEngine. It allows for printing errors to the render window

utils/path_util.py
jdent02: Probably not. There’s a way to retrieve the version number through the Python API

utils/cycles_shader_parsers.py
jdent02: Yes

dictoon: No idea about the consequences of doing so :)

properties/nodes.py
dictoon: Gotcha.

utils/path_util.py
dictoon: I'd say this needs to be changed before it can be merged. Also, Linux. Maybe there's a better, cross-platform way to do this?

utils/cycles_shader_parsers.py
jdent02: Should be fine.  Blender includes NumPy with it's Python interpreter

utils/cycles_shader_parsers.py
jdent02: Yep.

utils/path_util.py
jdent02: Path is the same for Linux so this will work for all platforms

Issue #445, opened 2019-09-04
Ignore OSL shaders that don't have the right attributes


dictoon:
I have an `APPLESEED_SEARCHPATH` environment variable on my system that points to appleseed's shader directories. That caused blenderseed to find 3ds Max shaders (among others) and it crashed because the `as_max_color_texture` shader has a parameter `U` of type `float that hasn't got a `default` attribute:

![image](https://user-images.githubusercontent.com/321290/64288372-3a196c00-cf62-11e9-868d-fab286279914.png)

Temporarily removing the `APPLESEED_SEARCHPATH` environment variable fixed the problem but it would be nice not to have to do that.

jdent02:
This fix isn't quite as easy as it seems.  The addon can't make any assumptions about what's in APPLESEED_SEARCHPATH since there's no certainty that it's organized in a particular way (like the appleseed shader directory).  I suppose anything containing the terms '3DSMax' or 'Maya' could be excluded but that might remove perfectly usable OSL shaders that just have those words in their filepaths.  I think in this case we just have to trust the user to point that env variable at the right place.

dictoon:
What I meant is that instead of chocking or crashing when encountering a shader that blenderseed can't handle, it should simply mention that in the log and skip it.

dictoon:
We need to fix this problem and never crash, regardless of how incompatible OSL shaders encountered in the search paths are. Problematic shaders must be ignored but shader discovery must proceed without crashing.

dictoon:
Bumping this up, this one is a major issue for me.

dictoon:
> So, after looking at this a bit more I can't just ignore shaders that don't have defaults for some variables.

Why not?

My point is simply that blenderseed should survive and continue to work fine even if the user throws random, non-compliant OSL shaders in his search paths.

dictoon:
One way to achieve this would be to wrap the shader discovery/parsing code in a giant try-catch all block and simply skip any shader that causes an exception to be thrown.

PR #517, opened 2020-06-10
Fix for #445 (OSL float without discreet default)

jdent02:


utils/osl_utils.py
```diff
@@ -305,7 +305,7 @@ def update_sockets(self, context):
 
         elif param['type'] == "float":
 
-            kwargs = {'name': param['name'], 'description': helper, 'default': float(default)}
+            kwargs = {'name': param['name'], 'description': helper, 'default': float(default)} if 'default' in keys else {'name': param['name'], 'description': helper}
```
dictoon, line 308: Conditionally adding
```
kwargs['default'] = float(default)
```
when the `default` key is present would be easier to read, as you already do for other optional keys below.

jdent02:
Fixed your suggestion

PR #518, opened 2020-06-23
Add sun positioning algorithm.

mororo250:
Make https://github.com/appleseedhq/appleseed/pull/2870 works in blenderseed.

### DIFF
--- a/properties/world.py
+++ b/properties/world.py
@@ -28,7 +28,8 @@
 import bpy
 
 from ..utils import util
-
+from datetime import datetime
+from ..translators.world import update_sun_pos
 
 class AppleseedSSSSetsProps(bpy.types.PropertyGroup):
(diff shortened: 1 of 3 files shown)


properties/world.py
```diff
@@ -46,21 +47,93 @@ class AppleseedSSSSets(bpy.types.PropertyGroup):
...
+                                        step=5.0,
+                                        default=0.0,
+                                        update=update_sun_pos)
+                                           
+    longitude:  bpy.props.FloatProperty(name="Longitude",
+                                        min=-180, 
+                                        max=180, 
+                                        step=5.0,
+                                        default=0.0,
+                                        update=update_sun_pos)
```
jdent02, line 136: For the above parameters I would add in some kind of appleseed prefix, such as 'as_' in order to avoid potential conflict with another addon.  Such things have happened in the past.

PR #524, opened 2020-10-07
Add new post-processing stages

laurelkeys:
🚧 Work in progress 🚧

Addresses https://github.com/appleseedhq/blenderseed/issues/523


jdent02:
For what it's worth your patch looks good so far.

